PR #228, opened 2016-10-24
Proxy for MapConfigs

milafrerichs:
## What does this PR do?
Adds the ability to use a proxy for the map config sources.

It can be useful especially for testing to use a proxy instead of the direct source url.
A lot of geonode servers don't support CORS and can't be integrated and displayed without a proxy in between.
This change helps in adding support for a proxy. if no proxy is passed in, nothing changes.


bartvde:
Thanks @milafrerichs I will merge once Travis agrees

PR #236, opened 2016-10-31
Drag and drop support for LayerList

### LINKED ISSUE #234: drag and drop support for reorder
Let's see if we can replace the buttons with drag and drop

bartvde:
closes #234


bartvde:
Haven't found a good way to solve the unit testing issue as yet.

bartvde:
The only way seems to be a test transform

PR #238, opened 2016-11-09
Changing to react-table

bartvde:
SDK-244


bartvde:
Just realised not all data has a fid, e.g. the GeoJSON exported from QGIS ....

PR #247, opened 2016-11-24
fix layers remove from inside groups

simod:


bartvde:
looks perfect @simod thank you for this contribution

PR #325, opened 2017-01-09
added a file watcher and parallelshell to support watching .js and .css

jjmulenex:
Added Parallelshell to support a double watch. 
Added the ability to watch css and JS and move them to the release/
setup: npm i
run: npm run watch:all
also added npm watch:css just to watch for CSS changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,9 @@
   "scripts": {
     "build:babel": "NODE_ENV=release babel ./js --out-dir ./release",
     "start": "babel --watch ./js --out-dir ./release",
+    "watch:all": "parallelshell 'npm run start' 'npm run watch:css'",
+    "watch:css": "onchange 'js/components/*.css' -- npm run build:copy:css",
+    "babel": "babel --watch ./js --out-dir ./release",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -11,6 +11,9 @@
   "scripts": {
     "build:babel": "NODE_ENV=release babel ./js --out-dir ./release",
     "start": "babel --watch ./js --out-dir ./release",
+    "watch:all": "parallelshell 'npm run start' 'npm run watch:css'",
+    "watch:css": "onchange 'js/components/*.css' -- npm run build:copy:css",
+    "babel": "babel --watch ./js --out-dir ./release",
```
bartvde: this seems to be the same as start?

bartvde: maybe your intention was to replace start with watch:all and hence you renamed  the current start command to babel

just guessing here

I'd be in favour of having npm start run watch:all, i.e. so rename watch:all to start in here

jjmulenex: I didnt know if there was a use case to have just .js watched so I left it in place. I can move it into start if that is NOT the case

jjmulenex: il make the change and push

PR #326, opened 2017-01-10
Theme fixes

bartvde:
SDK-295

as reported by @simod

### DIFF
--- a/__tests__/components/AddLayer.test.js
+++ b/__tests__/components/AddLayer.test.js
@@ -8,6 +8,8 @@ import raf from 'raf';
 import ol from 'openlayers';
 import intl from '../mock-i18n';
 import injectTapEventPlugin from 'react-tap-event-plugin';
+import getMuiTheme from 'material-ui/styles/getMuiTheme';
+import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
 import AddLayer from '../../js/components/AddLayer';
 
 raf.polyfill();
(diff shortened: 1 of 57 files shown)


bartvde:
so the issue here is that the test cases fail, that's probably why I originally made those changes in the components.

bartvde:
I need to find out how I can modify the test cases instead.

bartvde:
selection_009

bartvde:
For test cases that don't use the return from ReactDOM.render this is a solution, but not otherwise:

```
diff --git a/__tests__/components/AddLayer.test.js b/__tests__/components/AddLayer.test.js
index d9caae1..ad80cbd 100644
--- a/__tests__/components/AddLayer.test.js
+++ b/__tests__/components/AddLayer.test.js
@@ -8,6 +8,8 @@ import raf from 'raf';
 import ol from 'openlayers';
 import injectTapEventPlugin from 'react-tap-event-plugin';
+import getMuiTheme from 'material-ui/styles/getMuiTheme';
+import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
 import AddLayer from '../../js/components/AddLayer';
 
 raf.polyfill();
@@ -66,7 +68,9 @@ describe('AddLayer', function() {
   it('generates the correct drop zone', function(done) {
     var container = document.createElement('div');
     ReactDOM.render((
-      <AddLayer intl={intl} map={map}/>
+      <MuiThemeProvider muiTheme={getMuiTheme()}>      
+        <AddLayer intl={intl} map={map}/>
+      </MuiThemeProvider>
     ), container);
     var button = container.querySelector('button');
     TestUtils.SimulateNative.touchStart(button, tapDataInjector(0, 0));
```

PR #362, opened 2017-01-25
Nuka carousel

wnordmann:
Adding Nuka Carousel in to the bookmark component


js/components/Bookmarks.css
```diff
@@ -1,19 +1,18 @@
 .slick-prev:before, .slick-next:before {
```
bartvde: I guess this can be removed as well?

js/components/Bookmarks.js
```diff
@@ -15,19 +15,24 @@ import ol from 'openlayers';
 import Slider from 'react-slick';
 import IconMenu from 'material-ui/IconMenu';
 import MenuItem from 'material-ui/MenuItem';
+import SvgIcon from 'material-ui/SvgIcon';
```
bartvde: import not used?

js/components/Bookmarks.js
```diff
@@ -15,19 +15,24 @@ import ol from 'openlayers';
...
 import pureRender from 'pure-render-decorator';
 
 const messages = defineMessages({
-  dropdowntext: {
-    id: 'bookmarks.dropdowntext',
-    description: 'Text to use on the Bookmarks drop down',
-    defaultMessage: 'Bookmarks'
-  }
+    dropdowntext: {
+        id: 'bookmarks.dropdowntext',
```
bartvde: 2 spaces is the convention

js/components/Bookmarks.js
```diff
@@ -40,192 +45,298 @@ const messages = defineMessages({
...
-    if (this._resolution === undefined) {
-      view.once('change:resolution', function(evt) {
-        this._resolution = evt.target.getResolution();
-      }, this);
-    }
-    this.state = {
-      value: null
+      speed: React.PropTypes.number,
+      /**
+      *   Animation duration.
```
bartvde: this comment doesn't seem to match the key?

js/components/Bookmarks.js
```diff
@@ -40,192 +45,298 @@ const messages = defineMessages({
...
-      });
-      carouselChildren.unshift(<div key='intro'><h2>{this.props.introTitle}</h2><p>{this.props.introDescription}</p></div>);
-      return (
-        <div className={classNames('sdk-component story-panel', this.props.className)}>
-          <Slider {...this.props} arrows={true} afterChange={this._afterChange.bind(this)}>
-            {carouselChildren}
-          </Slider>
-        </div>
-      );
+    dots = React.createClass({
```
bartvde: don't use React.createClass, we are trying to write in ES6

bartvde: also it might be best to only have a single class per file, we might even have an eslint rule to enforce this

js/components/Bookmarks.js
```diff
@@ -40,192 +45,298 @@ const messages = defineMessages({
...
+        displayName: 'dots',
+
+        render: function render() {
+            var self = this;
+            var indexes = this.getIndexes(self.props.slideCount, self.props.slidesToScroll);
+            return React.createElement('ul', {
```
bartvde: please use JSX here, like elsewhere in the sdk

bartvde:
Also please look at what can be removed from here: https://github.com/boundlessgeo/sdk/blob/master/scripts/install.js#L14

js/components/Bookmarks.js
wnordmann: I agree it's vauge but it's the wording from the nuka-carousel code

js/components/Bookmarks.js
wnordmann: correct, good catch

PR #367, opened 2017-01-27
Fix up url handling for RESTService

bartvde:
we also need a way to use the access token in the requests


bartvde:
I guess we should introduce a restParams config next to restUrl, and then set that from geonode

bartvde:
The whole restUrl stuff can be removed we will only use the geoserver source instead, since the geonode gs/proxy doesn't work with the new auth system

PR #371, opened 2017-02-01
Callbacks to promises

wnordmann:
First example of converting from callbacks to promises

### DIFF
--- a/js/components/AddLayerModal.js
+++ b/js/components/AddLayerModal.js
@@ -181,9 +181,7 @@ class AddLayerModal extends React.PureComponent {
     return {muiTheme: this._muiTheme};
   }
   componentWillUnmount() {
-    if (this._request) {
-      this._request.abort();
-    }
+
   }
   static addNewTypes = [
     'WMS',
@@ -205,7 +203,6 @@ class AddLayerModal extends React.PureComponent {
     var me = this;
     const {formatMessage} = this.props.intl;
     var failureCb = function(xmlhttp) {
-      delete me._request;
       if (!xmlhttp || xmlhttp.status === 0) {
         me._setError(formatMessage(messages.corserror));
       } else {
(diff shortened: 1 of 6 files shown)


js/components/AddLayerModal.js
```diff
@@ -202,25 +200,22 @@ class AddLayerModal extends React.PureComponent {
     var source = this.state.sources[this.state.source];
     var url = source.url;
     var service = AddLayerModal.services[source.type];
-    var me = this;
+    var self = this;
+
+    service.getCapabilities(url)
```
bartvde: we do not to keep passing the opt_proxy param here

js/components/AddLayerModal.js
```diff
@@ -181,9 +181,7 @@ class AddLayerModal extends React.PureComponent {
     return {muiTheme: this._muiTheme};
   }
   componentWillUnmount() {
-    if (this._request) {
-      this._request.abort();
```
bartvde: how will this be handled with promises? Can we even abort the request when the component gets unmounted before the success "callback"  gets called to prevent setting state on an unmounted component?

js/components/AddLayerModal.js
```diff
@@ -202,25 +200,22 @@ class AddLayerModal extends React.PureComponent {
...
     const {formatMessage} = this.props.intl;
-    var failureCb = function(xmlhttp) {
-      delete me._request;
-      if (!xmlhttp || xmlhttp.status === 0) {
-        me._setError(formatMessage(messages.corserror));
-      } else {
-        me._setError(xmlhttp.status + ' ' + xmlhttp.statusText);
-      }
-      if (onFailure) {
-        onFailure();
```
bartvde: how is this handled with the promises?  Since when we refresh the service we call getCaps with another failure callback:

```
  _refreshService(onFailure) {
    this._getCaps(onFailure);
  }
```

js/services/WMSService.js
```diff
@@ -39,13 +39,13 @@ class WMSService {
...
-  getCapabilities(url, onSuccess, onFailure, opt_proxy) {
-    return util.doGET(this.getCapabilitiesUrl(url, opt_proxy), function(xmlhttp) {
-      var info = wmsCapsFormat.read(xmlhttp.responseText);
-      onSuccess.call(this, info.Capability.Layer, this._getGetMapUrl(info));
-    }, function(xmlhttp) {
-      onFailure.call(this, xmlhttp);
-    }, this);
+  getCapabilities(url) {
+    var self = this;
+    return util.fetchGet(this.getCapabilitiesUrl(url)).then(function(xmlhttp) {
```
bartvde: we need the proxy for geonode-client here still

js/components/AddLayerModal.js
wnordmann: I'm pretty sure this is better addressed with the redux, the fetch will be moved to a dispatch and an abort dispatch can be setup.  I'd like to neglect this functionality for the callbackToPromise store and revisit in the redux story

js/components/AddLayerModal.js
wnordmann: We can still pass functions around, I think the refresh on failure can be setup with redux and monitoring state.

js/components/AddLayerModal.js
bartvde: Make sense, I only add it back then because I had failing unit tests.

bartvde:
@wnordmann so maybe we should put this on hold until we finish the redux migration you think? Or doesn't that make sense?

PR #388, opened 2017-02-17
Inline dialog

bartvde:
SDK-312

selection_091


bartvde:
selection_094

bartvde:
I'm going to merge this though there are still some issues left, will tackle those separately

PR #397, opened 2017-02-23
Add a BCS Service

bartvde:
cc @thirystheory 

I changed it a bit, I will also update the sdk-apps example soon

Is there a known issue with the last base map, I'm getting exceptions on that one.

selection_104


thirystheory:
Yes there is a know issue with the digital globe basemap, recent imagery we
are working thru that on the API end of things.

I will check out when you all changed and ask any questions.

Thank you.
Liz

PR #403, opened 2017-03-03
New filter editor

bartvde:
SDK-321
SDK-314

 basically replicates what maputnik, except for the NOR (only AND and OR for now)


bartvde:
If we are going to have the MapBox format as the intermediate one, we need to adapt the SLDService to generate those instead.

PR #408, opened 2017-03-07
Structure LayerStore tests differently

bartvde:
SDK-287

we won't know if this fixes it until we have run Travis a few times


bartvde:
I restarted the build 3 times and did not get the failure, so fingers crossed this truly fixes it. But we won't know for a while.

PR #412, opened 2017-03-08
New table layout

bartvde:
SDK-350

selection_124


bartvde:
Closing this one because of the subdirectory change, will open up a new one

I think we should just merge this since it's a big improvement wrt space over the current behaviour, we can always refine later

PR #432, opened 2017-03-28
Fix conflicting react versions.

theduckylittle:
Material UI was including React at 15.4.2, but the
application's React version was set to 15.3.2.  This was
causing all of the tests to fail due to multiple included
versions of React.

This change puts all of the versions into harmony.  At least,
they are harmonious enough for all of the tests to pass in master.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,9 +55,9 @@
   "author": "Boundless Spatial, Inc.",
   "license": "Apache-2.0",
   "peerDependencies": {
-    "react": "15.3.2",
-    "react-dom": "15.3.2",
-    "react-tap-event-plugin": "^1.0.0"
+    "react": "15.4.2",
+    "react-dom": "15.4.2",
+    "react-tap-event-plugin": "^2.0.0"
   },
   "dependencies": {
     "blueimp-canvas-to-blob": "2.2.2",
@@ -129,11 +129,11 @@
     "phantomjs-polyfill-object-assign": "0.0.2",
     "phantomjs-prebuilt": "^2.1.7",
     "raf": "^3.2.0",
-    "react": "15.3.2",
-    "react-addons-test-utils": "15.3.2",
+    "react": "15.4.2",
+    "react-addons-test-utils": "15.4.2",
     "react-docgen": "^2.2.0",
-    "react-dom": "15.3.2",
-    "react-tap-event-plugin": "^1.0.0",
+    "react-dom": "15.4.2",
+    "react-tap-event-plugin": "^2.0.0",
     "rimraf": "^2.5.3",
     "sinon": "^1.12.2",
     "through": "^2.3.8",


bartvde:
we don't want to go to 15.4.2 there are issues with npm link

bartvde:
I'd recommend setting up apps like we do in here: https://github.com/boundlessgeo/sdk-generator/blob/master/bin/web-sdk#L93:L98

bartvde:
material-ui 0.16.4 has ^15.0.0 for react so I wonder why this would be incompatible? https://github.com/callemall/material-ui/blob/v0.16.4/package.json#L42

bartvde:
@theduckylittle can you explain a bit more how you ran into this? All tests have been passing on master for us for a long time

theduckylittle:
Not a problem, the explanation is fairly simple!

Forked the repository this morning.  So a fresh clone should be able to recreate the issue.

Executed:
  npm install
  npm build
  npm test

Every test failed.  I removed node-modules then upgraded the packages as seen in this pull request.  Then I repeated the steps above and every test passed.   

I suspect the reason why everyone else's tests are passing is because they've had material-ui installed for quite a while and that instance of material-ui is using React 15.3.2.

bartvde:
what node/npm versions are you on?
I'm trying now with a fresh run.
Make sure you're on LTS for node/npm

theduckylittle:
Clean clone, then running npm install creates the following issue:
```
ducky@ThinkDuck ~/boundless/sdk $ grep version node_modules/material-ui/node_modules/react/package.json
  "version": "15.4.2",
ducky@ThinkDuck ~/boundless/sdk $ grep version node_modules/react/package.json 
  "version": "15.3.2",
```

bartvde:
seems you're on an ancient version of node/npm? Npm has 1 level node_modules for a while already

theduckylittle:
```
ducky@ThinkDuck ~/boundless/sdk $ npm --version
2.15.11
ducky@ThinkDuck ~/boundless/sdk $ node --version
v4.8.1
```

bartvde:
yeah way too old, use node 6

bartvde:
which reminds me I need to add a section to our package.json
```
  "engines": {
    "node": ">6"
  },
```

bartvde:
I'll add it to the README as well, it's in the docs but it should be in the README as well.

https://boundlessgeo.github.io/sdk/book/getting_started.html

theduckylittle:
I'm not confident that will be the fix unless Node 6 has specifically added magic to fix library conflict resolutions.  I'm working on upgrading now.

bartvde:
it has

bartvde:
https://docs.npmjs.com/how-npm-works/npm3

theduckylittle:
npm3 does fix the version conflicts. Closing.

PR #435, opened 2017-03-29
Simplify Chart component and improve UI/UX

bartvde:
SDK-335


bartvde:
The closer is now part of the component, with an onClose callback function

selection_172

PR #438, opened 2017-04-04
New UI/UX for WFST and Edit

bartvde:
SDK-384


bartvde:
The EditForm can now become the form without the actual buttons, the buttons can be moved to the popup instead

bartvde:
We can get rid of the whole EditForm abstraction actually

bartvde:
the layer combo box won't work if stopEvent is not false, but this leads to other issues

bartvde:
Another issue I have is: the layer select combo's filter function only gets called if the LayerStore changes. In this case I want to limit the list based on the geometry type but this does not get called when it's needed. Need to find a solution to this problem.

bartvde:
Last issue resolved with: https://github.com/boundlessgeo/sdk/pull/438/commits/569e5ee027356d92233d8dbcd8c516dc178bda53

bartvde:
I'm going to merge this, it's largely done. We can improve it later.

PR #439, opened 2017-04-04
Left nav

wnordmann:
Reworked cherry picks to make a clean branch


bartvde:
is this ready for review @wnordmann ? Are any changes to apps necessary?

src/components/LeftNav.js
```diff
@@ -0,0 +1,163 @@
...
+       */
+      children: React.PropTypes.node,
+      /**
+       * array of <tab> components to be added to <Tabs>
+       */
+      tabList: React.PropTypes.node,
+      /**
+       * NavItem components to be added to LeftNav
+       */
+      navitems: React.PropTypes.node,
```
bartvde: should we do navItems instead since we have tabList?

src/components/LeftNav.js
```diff
@@ -0,0 +1,163 @@
...
+       * If true drawer is opened
+       */
+      open: React.PropTypes.bool,
+      /**
+       *  Callback for closing the drawer
+       */
+      onRequestClose: React.PropTypes.func
+
+    };
+    constructor(props, context) {
```
bartvde: since it's just the super call, it can go right this constructor?

src/components/LeftNav.js
```diff
@@ -0,0 +1,163 @@
...
+      super(props);
+    }
+    componentDidMount() {
+      if (this.props.tabList) {
```
bartvde: should we check for length > 0 as well?

src/components/LeftNav.js
```diff
@@ -0,0 +1,163 @@
...
+    static childContextTypes = {
+      muiTheme: React.PropTypes.object.isRequired
+    };
+    state = {
+      menuOpen: false
+    };
+
+    handleToggle = () => {
+      this.setState({
+        menuOpen: !this.state.open
```
bartvde: open or menuOpen? Is there open in state?

src/components/LeftNav.js
```diff
@@ -0,0 +1,163 @@
...
+          }else {
+            icons[tab.props.value] = {icon: <span/>, onTouchTap: null};
+          }
+          menuItems.push(<MenuItem primaryText = { tab.props.label } value = { tab.props.value }/>);
+        });
+      }
+      this.rightIcon = icons;
+      return (
+        <Drawer width = { this.props.width } open = { this.props.open } >
+          <AppBar title = { <span><span onTouchTap = { this.handleMenuTouchTap }> {this.state.menuText} <NavigationArrowDropDown style = { iconStyles }/></span>
```
bartvde: does this need a bind(this) ?

src/components/LeftNav.js
```diff
@@ -0,0 +1,163 @@
...
+            <Popover
+              open = { this.state.menuOpen }
+              anchorEl = { this.state.anchorEl }
+              anchorOrigin = {{horizontal: 'left', vertical: 'bottom'}}
+              targetOrigin = {{horizontal: 'left', vertical: 'top'}}
+              onRequestClose = {this.handleMenuRequestClose} >
```
bartvde: does this need a bind(this) ?

src/components/LeftNav.js
```diff
@@ -0,0 +1,163 @@
...
+                  <Menu onItemTouchTap = { this.handleMenuChange } children={menuItems}/>
```
bartvde: does this need a bind(this) ?

src/components/LeftNav.js
```diff
@@ -0,0 +1,163 @@
...
+               </Popover></span>}
+            iconElementLeft = { <IconButton> <NavigationArrowBack/> </IconButton>}
+            onLeftIconButtonTouchTap={this.close}
```
bartvde: does this need a bind(this) ?

src/components/StyleModal.js
```diff
@@ -84,6 +88,10 @@ class StyleModal extends React.PureComponent {
      */
     className: React.PropTypes.string,
     /**
+    * Drawer where true, modal when false, default false
+    */
+    isDrawer : React.PropTypes.bool,
```
bartvde: seems we now have a few components that share this isDrawer behaviour where they can switch between a Drawer and a Dialog, would be nice to have this code shared in a base class, similar to how I did the Dialog class. But this can come later.

src/components/StyleModal.js
```diff
@@ -302,18 +310,51 @@ class StyleModal extends React.PureComponent {
...
-        </Dialog>
+        </div>
+        <div className='footerButtons'>
+          {actions}
+        </div>
+      </Drawer>
+    );
+    var dialog = (
+      <Dialog ref='dialog'
+        inline={this.props.inline}
```
bartvde: I'm guessing that the whole drawer stuff is actually a replacement for the inline dialogs that I did right? So that prop can probably go, and the intermediate class could go as well. But this can be done later in refactor round.

src/components/LeftNav.js
wnordmann: No there arn't nav items, I was considering making a navitems type but I could manage with <tab> components.  It may be worth a re-work in the future.  I will removed the prop

src/components/LeftNav.js
wnordmann: Toggle is never used anyway
I'll remove it

src/components/LeftNav.js
wnordmann: I think I only need the .bind(this) on the `onLeftIconButtonTouchTap={this.close}` because `this.close` is using a passing in callback.  The other functions are internal to LeftNav.js it self.  If more callback were added more `.bind(this)` would be needed.  There is a small performance hit using `.bind(this)`

src/components/AddLayerModal.js
```diff
@@ -604,17 +619,33 @@ class AddLayerModal extends React.PureComponent {
...
+              onLeftIconButtonTouchTap={this.close.bind(this)}/>
+            <div className="noBorderPaper">
+              {select}
+              {content}
+            </div>
+            <div className='footerButtons'>
+              {actions}
+            </div>
+          </Drawer>);
+    var dialog = (<Dialog bodyStyle={{padding: 20}} inline={this.props.inline} className={classNames('sdk-component add-layer-modal', this.props.className)} actions={actions} autoScrollBodyContent={true} modal={true} title={formatMessage(messages.title)} open={this.props.open} onRequestClose={this.close.bind(this)}>
```
wnordmann: Correct, Inline is not needed it was a carry over from a copy/paste

src/components/LeftNav.js
bartvde: Please check what this is in the debugger, I thought with ES6 style React classes bind is always needed if the function uses ```this``` at all

src/components/Header.js
```diff
@@ -0,0 +1,89 @@
...
+    super(props);
+  }
+
+  state = {
+    menuOpen: false
+  };
+
+  handleToggle = () => {
+    this.setState({
+      menuOpen: !this.state.open
```
wnordmann: Again not used

src/components/StyleModal.js
wnordmann: Good idea

src/components/StyleModal.js
wnordmann: agreed.

PR #443, opened 2017-04-11
Use mapskin from npm instead

bartvde:


bartvde:
@wnordmann as a follow-up, we should probably untangle the import from layerlist. In the, I admit unlikely, scenario that someone does not have a LayerList, all icons would not work, right? Maybe we need to re-instate the main css file again in SDK.

PR #454, opened 2017-04-20
Table uiux followup

bartvde:
This came up in SDK-422 I noticed some of the changes the table were still pending, see also SDK-310

Also need to see if the paging toolbar and the header columns can always be visible, so WIP.


bartvde:
For the scrollbars, there is: https://github.com/tannerlinsley/react-table/wiki/Fixed-headers-with-scrollable-table-body but I cannot seem to get it to work properly

PR #455, opened 2017-04-20
Bring back toggle mode for layer list

bartvde:
This brings back toggle mode for layer list with a bit of css at the app level as can be seen here: https://github.com/boundlessgeo/sdk-apps/commit/ac9e312290b7afce9cd30a07ddc330d56235c8c2

@volaya we probably need to make a similar change as the one above in WAB

selection_232
selection_231

### DIFF
--- a/src/components/LayerList.js
+++ b/src/components/LayerList.js
@@ -235,6 +235,7 @@ class LayerList extends React.PureComponent {
     super(props);
     LayerStore.bindMap(this.props.map);
     this.state = {
+      visible: props.showOnStart,
       addLayerOpen: false,
       muiTheme: context.muiTheme || getMuiTheme(),
       baseLayer: ''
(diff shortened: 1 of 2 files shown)


bartvde:
For apps like QuickView we need to set showOnStart to true

bartvde:
@jjmulenex your latest commit broke the build, I just fixed it with: https://github.com/boundlessgeo/sdk/pull/455/commits/3c77dfd53ddec69785851cfdbde7fecbd76d31b4

bartvde:
Also, to prevent merge commits such as: https://github.com/boundlessgeo/sdk/commit/3ebb6f1708374e9602f799889fbd5e450c550778 see answers such as: http://stackoverflow.com/questions/30052104/how-to-avoid-merge-commits-from-git-pull-from-pushing-to-remote

PR #459, opened 2017-04-24
Playback

jjmulenex:
updated the playback control to fit the new creative.


src/components/Playback.js
```diff
@@ -61,6 +62,10 @@ class Playback extends React.PureComponent {
      */
     className: React.PropTypes.string,
     /**
+     * Title of the playback
+     */
+    Title: React.PropTypes.string,
+    /**
```
bartvde: this should be all lowercase

src/components/Playback.js
```diff
@@ -230,30 +235,60 @@ class Playback extends React.PureComponent {
...
       buttonIcon = <PauseIcon />;
     }
-    var controls;
+    var playButton;
+    var slider;
+    var datePicker;
+    var title;
+
+    if (this.props.title) {
+      title = (this.props.title);
```
bartvde: no need for brackets here

src/components/Playback.js
```diff
@@ -230,30 +235,60 @@ class Playback extends React.PureComponent {
...
+    } else {
+      title = ('Playback');
```
bartvde: same here, also I would put this in i18n messages

PR #479, opened 2017-05-02
Infopop attr filter

CarlyLanglois:


bartvde:
So if I understand the PR correctly, in addition to what we have for popupInfo:

 * string based template with [foo] for attribute foo
 * '#AllAttributes' to show all attributes

we are adding a third mode with an array of attributes to show right?

bartvde:
No the deletion breaks the basic example. Compare before and after:
selection_248
selection_249

CarlyLanglois:
thanks Bart. I've put the block back in that handles this type of popupInfo and changed where to use getGeometryName()

bartvde:
next step would be to document all the options somewhere but this can be done in a new PR

PR #497, opened 2017-05-15
Use stopEvent true on EditPopup

bartvde:
SDK-484


bartvde:
This breaks the combo box in the popup, sigh.

bartvde:
I'm going to revert this

PR #508, opened 2017-05-19
AddLayerModal testing

CarlyLanglois:


src/components/Chart.js
```diff
@@ -75,14 +75,20 @@ class Chart extends React.Component {
     /**
      * An array of configuration objects. Configuration objects have title, categoryField, layer,
      * valueFields, displayMode and operation keys.
+
```
wnordmann: Maybe some screen shots could be added to a more rich help doc

__tests__/components/Chart.test.js
```diff
@@ -58,11 +67,8 @@ describe('Chart', function() {
 
   it('generates the correct combo', function() {
     var container = document.createElement('div');
-    var onClose = function() {};
     ReactDOM.render((
-      <MuiThemeProvider muiTheme={getMuiTheme()}>
```
wnordmann: Do we not need muiTheme tested?

CarlyLanglois: i dont think so.. it kind of impedes what can be tested. I think the only reason it was here is because the component requires it, so I modified the component's context.

PR #514, opened 2017-05-24
Get table page size in a more dynamic way

bartvde:
Row height etc. might depend on font etc.


bartvde:
this kills the table when it's visible immediately

bartvde:
this is kind of a chicken egg problem

bartvde:
ok see: https://github.com/boundlessgeo/sdk/commit/e4df81008f419866d44dbcf8a315a19fd85c14d7

PR #515, opened 2017-05-25
Nocov test

CarlyLanglois:
starting to add ignore statements to assignments not necessary to test


bartvde:
interesting, didn't know about this concept. Would make sense to me to use them or not?

PR #522, opened 2017-06-02
Adding in modal close

wnordmann:
adding in zoomtom functions

### DIFF
--- a/src/components/GeocodingResults.js
+++ b/src/components/GeocodingResults.js
@@ -14,6 +14,7 @@
  import {connect} from 'react-redux';
  import Popover from 'material-ui/Popover';
  import Menu from 'material-ui/Menu';
+ import MenuItem from 'material-ui/MenuItem';
  import ol from 'openlayers';
  import {defineMessages, injectIntl, intlShape} from 'react-intl';
  import classNames from 'classnames';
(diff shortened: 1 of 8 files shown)


src/reducers/geocoding.js
```diff
@@ -3,19 +3,20 @@ export default (state = [], action) => {
...
       return {
         geocodingSearchString: action.text,
-        searching:true,
         showGeocodingResults:false
       }
     case 'GEOCODING_SELECT':
       return {
-        latitude:10,
-        longitude:10
+        showGeocodingResults:false
```
CarlyLanglois: also, we should probably return ...state, before the new results to not erase existing state object

src/components/GeocodingResults.js
```diff
@@ -127,9 +188,8 @@
  // Maps actions to props
  const mapDispatchToProps = (dispatch) => {
    return {
-   // You can now say this.props.createBook
-     // geocodingSearch: search => dispatch(geocodingActions.fetchGeocode(search))
-     closeGeocoding: search => dispatch(geocodingActions.fetchGeocode(search))
+     geocodingClose: search => dispatch(geocodingActions.geocodingClose()),
```
bartvde: is search still needed here?

PR #525, opened 2017-06-06
Sdk redux

bartvde:
Just opening up this PR already so I can get a better overview of the sdk-redux branch


bartvde:
it would be good to also make the split into components and containers in this branch

https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0

src/reducers/map.js
```diff
@@ -0,0 +1,119 @@
...
+          bounds: action.bounds
+        }
+      }
+    case  MAP.SET_VIEW:
+      const new_view = state.view;
+
+      // if the action specifies a projection when setting the view,
+      //  then pull that out and reproject the center.
+      if (action.projection) {
+        action.center = projectPoint(center, action.projection, state.config.projection);
```
CarlyLanglois: where does the center value come from here?

PR #529, opened 2017-06-12
only saving relevant zoom info in state tree

CarlyLanglois:
simplifying what is stored in state tree until more actions are created. trying to prevent erroneous, out-of-date map data from being stored in state tree

### DIFF
--- a/src/components/MapView.js
+++ b/src/components/MapView.js
@@ -14,16 +14,8 @@
  import ReactDOM from 'react-dom';
  import ol from 'openlayers';
  import './MapPanel.css';
- import {defineMessages, injectIntl, intlShape} from 'react-intl';
+ import {injectIntl, intlShape} from 'react-intl';
  import classNames from 'classnames';
-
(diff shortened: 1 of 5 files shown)


src/components/MapView.js
```diff
@@ -74,7 +66,7 @@
     this._proxy = context.proxy;
     this._requestHeaders = context.requestHeaders;
     if (this.props.hasOwnProperty('getMap')) {
-      this.props.getMap(this.props.map);
+      //this.props.getMap(this.props.map);
```
wnordmann: best not leave commented code in the final product

src/components/MapView.js
```diff
@@ -106,7 +97,7 @@
   }
   render() {
     return (
-      <div style={this.props.style} id={this.props.id} ref='map' className={classNames('sdk-component map-panel willie', this.props.className)}>
```
wnordmann: example of why 'test' should be the only values used when testing strange behavior

PR #530, opened 2017-06-12
Port the Home Button to Redux

theduckylittle:
1. BoundlessSdk will now accept a store as a "prop".  This is especially important for testing.  Otherwise it was difficult to get at a "clean" store for testing and investigating state changes.
2. The home button test works but we may need a cleaner way to handle the injection of tap-touch events.  I wrapped the injection in a try/catch but needing to repeat that in every test may get tiresome.
3. There's another small issue we will need to handle at somepoint.  The previous Home button used a bounding-box (extent prop) to define the home are but the current setView function only accepts a center and view.  The new version uses a center prop and a zoom prop to allow setting the "home".  Otherwise, it will listen for the first state change and consider that the home location.


bartvde:
@theduckylittle wrt the tap event in the tests, I think an approach like outlined here would be better: https://github.com/zeit/next.js/issues/232#issuecomment-268730651

I just tested it and it works good, the bundler will ensure only a single entry

This has been awkward in testing for a while, requiring the full test suite to be run, and running an individual test to be problematic. Thanks for triggering this and reminding me ;-)

```
bartvde@bartvde-XPS-13-9360:~/opengeo/git/sdk$ git diff
diff --git a/__tests__/components/AddLayerModal.test.js b/__tests__/components/AddLayerModal.test.js
index 5df347e..0a372d3 100644
--- a/__tests__/components/AddLayerModal.test.js
+++ b/__tests__/components/AddLayerModal.test.js
@@ -6,12 +6,11 @@ import {assert} from 'chai';
 import raf from 'raf';
 import ol from 'openlayers';
-import injectTapEventPlugin from 'react-tap-event-plugin';
 import AddLayerModal from '../../src/components/AddLayerModal';
 import TestUtils from 'react-addons-test-utils';
+import '../tap_events';
 
 raf.polyfill();
-injectTapEventPlugin();
 
 describe('AddLayerModal', function() {
 
diff --git a/__tests__/components/ArcGISRestLegend.test.js b/__tests__/components/ArcGISRestLegend.test.js
index a20a363..06b212d 100644
--- a/__tests__/components/ArcGISRestLegend.test.js
+++ b/__tests__/components/ArcGISRestLegend.test.js
@@ -6,6 +6,7 @@ import {assert} from 'chai';
 import raf from 'raf';
 import ol from 'openlayers';
 import ArcGISRestLegend from '../../src/components/ArcGISRestLegend';
+import '../tap_events'
 
 raf.polyfill();
 
diff --git a/__tests__/components/LayerList.test.js b/__tests__/components/LayerList.test.js
index a1cd017..f04dda6 100644
--- a/__tests__/components/LayerList.test.js
+++ b/__tests__/components/LayerList.test.js
@@ -10,6 +10,7 @@ import 'phantomjs-polyfill-object-assign';
 import getMuiTheme from 'material-ui/styles/getMuiTheme';
 import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
 import LayerList from '../../src/components/LayerList';
+import '../tap_events'
 // import TestUtils from 'react-addons-test-utils';
 
 raf.polyfill();
```

theduckylittle:
Two updates:
1. Fixed the issue between actionTypes and ActionTypes.
2. Handled the tap-event injection more sanely.  I actually made a `polyfills.js` that is in the root directory for the tests.  Both HomeButton and AddLayerPanel imports it and it lets the tests run smoothly both together and individually.

bartvde:
test update looks sane, we should probably do this for all tests eventually

__tests__/components/HomeButton.test.js
```diff
@@ -2,65 +2,67 @@
 
 import React from 'react';
 import ReactDOM from 'react-dom';
-import {assert} from 'chai';
-import raf from 'raf';
+import {expect} from 'chai';
 import ol from 'openlayers';
-import intl from '../mock-i18n';
+//import Intl from '../mock-i18n';
```
bartvde: we can get rid of the comment here?

__tests__/components/HomeButton.test.js
```diff
@@ -2,65 +2,67 @@
...
-      done();
-    });
-  });
-
-  afterEach(function() {
-    map.setTarget(null);
-    document.body.removeChild(target);
+    // reset the store before each test.
+    // store = configureStore();
+    done();
```
bartvde: can we get rid of the commented code?

__tests__/components/HomeButton.test.js
```diff
@@ -2,65 +2,67 @@
...
-    assert.equal(map.getView().getCenter()[0], 0);
+
+    let view = store.getState().mapState.view;
+
+    expect(view.center).to.deep.equal(init_center);
+    expect(view.zoom).to.equal(init_zoom);
+
+    // check that the state has goen back t
+    // assert.equal(map.getView().getZoom(), 1);
+    // assert.equal(map.getView().getCenter()[0], 0);
```
bartvde: can we get rid of the commented code?

src/components/HomeButton.js
```diff
@@ -64,37 +67,67 @@ class HomeButton extends React.PureComponent {
...
-      }
+
+    // default the view to null
+    let view = null;
+
+    // the state can be seeded automatically or manually,
+    // when center and zoom are set then that is used for
+    // the default view.
+    if (props.center) {
+      this.state.view = {
```
bartvde: does this work? Shouldn't it be ```this.state = {};``` first? Or ```this.state = {view: {}};``` ?

bartvde: Or should it be view = {} instead since you assign it to this.state later?

src/components/HomeButton.js
```diff
@@ -64,37 +67,67 @@ class HomeButton extends React.PureComponent {
...
 
-export default injectIntl(HomeButton);
+const mapStateToProps = (state) => {
+  return {
+    map: state.mapState
+  }
+}
+
+const mapDispatchToProps = (dispatch) => {
+  return {
```
bartvde: this will be automatic right? No need for specifying mapDispatchToProps if I've understood it correctly, it will automatically pass on dispatch if not specified

theduckylittle: Yes, I intentionally left it in there assuming feedback would be one of two items:
a) It's automatic, quit it. In which case it's easy to remove the code.
b) Map the specific action. In which case, it's easy to flush out the stub.

I prefer just calling dispatch directly but others may prefer the semantic wrappers. :) So I went with option (a). in the fix

bartvde: Ok, that is fine by me ;-)

PR #534, opened 2017-06-15
Support vector tiles layers

bartvde:
fixes SDK-569


bartvde:
the change to WFSService is needed due to a small regression in openlayers see https://github.com/openlayers/openlayers/pull/6931 for the details

PR #537, opened 2017-06-19
Fixed home button to match new view handling.

theduckylittle:
After the changes to the map reducer, the Home button was broken, this fixes the component and updates the tests.  

A lot of the diff is cleaning up the linter warnings.


__tests__/components/HomeButton.test.js
```diff
@@ -1,22 +1,21 @@
...
+import {IntlProvider} from 'react-intl';
 
 import configureStore from '../../src/stores/Store';
-import { setView } from '../../src/actions/MapActions';
+import {setView} from '../../src/actions/MapActions';
 
 import BoundlessSdk from '../../src/components/BoundlessSdk';
 
-import polyfills from '../polyfills';
+import polyfills from '../polyfills'; // eslint-disable-line no-unused-vars
```
bartvde: can we do ```import '../polyfills';``` instead?

theduckylittle: Neat! That works. I've updated that test and the others that were importing 'polyfills'.

PR #538, opened 2017-06-19
Sdk redux bookmarks center

CarlyLanglois:
Bookmarks component uses setView action to move map to selected bookmark.
User-defined bookmarks must each have a center and resolution.


__tests__/components/Bookmarks.test.js
```diff
@@ -1,112 +1,100 @@
...
+import {expect} from 'chai';
+import {setView} from '../../src/actions/MapActions';
+import 'phantomjs-polyfill-object-assign';
+import ol from 'openlayers';
 import intl from '../mock-i18n';
-import {Bookmarks} from '../../src/components/Bookmarks';
+import Bookmarks from '../../src/components/Bookmarks';
+import BoundlessSdk from '../../src/components/BoundlessSdk';
+import configureStore from '../../src/stores/Store';
+import polyfills from '../polyfills'; // eslint-disable-line no-unused-vars
```
theduckylittle: @bartvde  has suggested changing this to "import '../polyfills" which removes the need for silencing the linter.  There should be some examples of that but I'm not sure if they're only in pull requests.  It's a small change but I'd rather we do it now than later.

__tests__/components/Rotate.test.js
```diff
@@ -15,8 +15,6 @@ import configureStore from '../../src/stores/Store';
 
 import polyfills from '../polyfills'; // eslint-disable-line no-unused-vars
```
theduckylittle: Same comment as above.

src/actions/MapActions.js
```diff
@@ -25,3 +25,9 @@ export function setView(view) {
     view
   }
 }
+export function extentToResolution(map) {
```
theduckylittle: I see this extentToResolution in the action-related files and in the MapConfigService.

1. There really is no need for a ConfigService with redux.  The configuration is the state.
2. I'm not seeing this used anywhere, so can it be eliminated?

CarlyLanglois: yeah it definitely can

PR #539, opened 2017-06-20
Sdk redux ZoomSlider

CarlyLanglois:
zoom slider using and updating redux resolution value


src/components/ZoomSliderView.js
```diff
@@ -28,54 +30,64 @@ class ZoomSlider extends React.PureComponent {
...
-  */
-  _handleChangeResolution() {
-    this.setState({
-      value: this._getValue(this.props.map.getView().getResolution())
-    })
-    //var olResolution = this.map.getView().getResolution();
-    //this.props.getResolutionValue(this._getValue(olResolution))
+
+  getResolutionFn() {
+    return (new ol.View({projection: this.props.map.config.projection})).getResolutionForValueFunction();
```
CarlyLanglois: a couple questions on this line - where is getResolutionForValueFunction defined? and is it intentional to put an openlayers view object in this component? I had been trying to avoid keeping the components dependent on ol

theduckylittle: It's defined here: https://github.com/openlayers/openlayers/blob/master/src/ol/view.js#L688

I'm not sure why it doesn't make it into the documentation.  It's also the function that the regular OpenLayers slider uses.

So the `ol.View` is really being used to do the calculations based on the projection. Between using it and trying to pass the map-object around I think this is the lesser of two evils.  We could also abstract this out.

bartvde: this is internal in OpenLayers, so not API, cannot be used

theduckylittle: So, @bartvde wins... this version still uses ol.View but for some reason I didn't see the getMax/MinResolution functions... sometimes it helps to slow down and read the code a little better! That said, the resolutions are coming from ol.View for now and the min/max are calculated when necessary.  We can look at storing these in the store but if they can be calculated that's an improvement over depending on passing around the ol.Map object.

src/components/ZoomSliderView.js
```diff
@@ -28,54 +30,69 @@ class ZoomSlider extends React.PureComponent {
...
-    })
-    //var olResolution = this.map.getView().getResolution();
-    //this.props.getResolutionValue(this._getValue(olResolution))
+
+  getResolutions() {
+    const v = new ol.View({projection: this.props.map.config.projection});
+    return {
+      min: v.getMinResolution(),
+      max: v.getMaxResolution()
+    }
```
CarlyLanglois: would it create any issues to take this out of zoomSlider and put it into the Map component that already handles the openLayer view? From what Im seeing, projection is set by default so max and min res are too?

theduckylittle: With React components are not meant to talk to each-other.  The Projection is not actually coming from the map component but it's being set in the state tree.  And it's just a string not the actual ol.View or ol.proj/proj4js projection.  As we move the code forward the Map component will need to start using the state-tree to define it's own settings (such as projection).

A logical next step would be to require the user to configure the BoundlessSdk component with the map information (such as Projection, resolutions, bound limits) and by setting the projection we could then populate the min/max resolution.  And when/if we do that, then we can just update those two lines to pull the resolutions from the state tree instead of calculating them.

bartvde: yeah, keep in mind that we might just configure the BoundlessSdk component with a MapBox style GL JSON doc in the future, which has all that state

CarlyLanglois: ok, got it. thanks

PR #541, opened 2017-06-20
Sdk redux zoom to lat lon

CarlyLanglois:
- Map component sets the map's projection in the view state
- ZoomToLatLon component uses this projection and given resolution and lat/lon values to set new view with setView action
- should the open/not open status of this dialog be tracked in state tree too?


CarlyLanglois:
ok this is up to date without using mapconfig. should be ready to merge. 
one more change - if wanted - would be to add the open/close dialog status into the state tree

PR #542, opened 2017-06-21
Start migrating "MapConfigService" to State Tree

theduckylittle:
The MapConfigService needs to be deprecated and replaced with a dedicated part of the state tree which holds the critical map information.

This is a start on that in order to support some of the work being done by @CarlyLanglois .


src/reducers/map.js
```diff
@@ -20,17 +20,52 @@ function convertFromExtent(mapSize, extent, rotation) {
...
+  const p = (new ol.geom.Point(pt)).transform(srcProj, destProj);
+  return p.getCoordinates();
+}
+
+const defaultState = {
+  config: {
+    // default projection for OpenLayers maps.
+    projection: 'EPSG:3857',
+    // bounds are specified in EPSG:4326 for interoperability.
+    bounds: [-180.0, -90.0, 180.0, 90.0]
```
bartvde: are we doing bounds or center and zoom?

src/reducers/map.js
```diff
@@ -20,17 +20,52 @@ function convertFromExtent(mapSize, extent, rotation) {
...
+        ...state,
+        config: {
+          ...state.config,
+          bounds: action.bounds
+        }
       }
     case  MAP.SET_VIEW:
       const new_view = state.view;
+
+      // if the action specifys a projection when setting the view,
```
bartvde: specifies

theduckylittle: Doh.

src/reducers/map.js
theduckylittle: Changed to maxExtent for clarity.  We're using center/resolution to determine the "view", this is meant to define the map's overall configuration and I was just stubbing in the ability to limit the map's maximum bounds.

theduckylittle:
With @bartvde 's approval, i'm merging this in since we're using the improvements in other components.

PR #545, opened 2017-06-27
Make sure set selected as filter works when filtered

bartvde:
Fixes https://github.com/boundlessgeo/qgis-webappbuilder-plugin/issues/415


bartvde:
But I do see an additional issue if you first select and then filter, it is also not correct see:

selection_320

bartvde:
Need to write some tests still

bartvde:
Not so easy to write an automated test for these changes, so I'm going to hold off on that for now

PR #546, opened 2017-07-04
Added support for geojson source + vector layers

theduckylittle:
1. Map component now adds vector layers to the map
   appropriately.

2. Package.json now includes the MapBox-to-OL style dependency
   for translating GL styles to OL styles.

3. Basic demo now has an "add random point" button to demonstrate
   the addFeatures action working dynamically.

### DIFF
--- a/examples/basic/app.js
+++ b/examples/basic/app.js
@@ -47,10 +47,10 @@ function main() {
       type: 'Feature',
       geometry: {
         type: 'Point',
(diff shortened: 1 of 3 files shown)


bartvde:
test failure fixed with: https://github.com/boundlessgeo/sdk/commit/21436a6f529dcb5f784a01e34d624568acd432a0

PR #547, opened 2017-07-05
Allow loading remote document for Mapbox GL Style JSON

bartvde:
This now allows for reloading a new context document


bartvde:
one thing I notice is that the null island point remains in the map, even though the source and layer are gone

theduckylittle:
I suspect that's a consequence of the component not updating the sources/layers properly to handle a "dropped" or missing layer.  I'll check it out.

src/action-types.js
```diff
@@ -8,6 +8,7 @@ export const MAP = {
   REMOVE_SOURCE: 'MAP_REMOVE_SOURCE',
   UPDATE_LAYER: 'MAP_UPDATE_LAYER',
   ADD_FEATURES: 'MAP_ADD_FEATURES',
-
   SET_LAYER_VISIBILITY: 'MAP_SET_LAYER_VISIBILITY',
+  REQUEST_CONTEXT: 'MAP_REQUEST_CONTEXT',
```
bartvde: Yeah it was more to show a spinner, but that can come later

src/components/map.js
```diff
@@ -170,49 +170,51 @@ export class Map extends React.Component {
...
       const layer = layersDef[i];
       const is_visible = layer.layout ? layer.layout.visibility !== 'none' : true;
 
       if(!(layer.id in this.layers)) {
-        const layer_src = sourcesDef[layer.source];
-        let new_layer = null;
-        if(layer_src.type === 'raster') {
-          if('tiles' in layer_src) {
-            new_layer = new TileLayer({
+        if (layer.type !== 'background') {
```
bartvde: it won't become more flat, since we will still have an else statement to get the source (common code for the else's): ```const layer_src = sourcesDef[layer.source];```

bartvde: I made the change though, but as I said it won't be more flat

src/actions/map.js
```diff
@@ -61,3 +63,31 @@ export function setLayerVisibility(layerId, visibility) {
...
+
+export function receiveContext(context) {
+  return {
+    type: MAP.RECEIVE_CONTEXT,
+    context
+  }
+}
+
+// thunk action creator
+export function fetchContext(url) {
```
bartvde: Sure that makes sense

bartvde:
I implemented your requested changes @theduckylittle 

I will add some tests now

PR #553, opened 2017-07-07
Added support for 'vector' type layers.

theduckylittle:
This does add support for the vector layers
but multiple layer per source is demonstrating a bug
which only renders the first layer.


src/components/map.js
```diff
@@ -25,6 +30,8 @@ import VectorSource from 'ol/source/vector';
 
 import GeoJsonFormat from 'ol/format/geojson';
 
+//_ol_layer_VectorTileRenderType_.VECTOR,
+import RENDER_MODE from 'ol/layer/vectortilerendertype';
```
bartvde: is this needed? I don't see it used.

PR #554, opened 2017-07-07
Add Support for Layer Titles

theduckylittle:
1. Moved _sourcesVersion, _layersVersion, and _dataVersion
   into the metadata object.  I used "bnd:" as the Boundless prefix
   to prevent clashing. (MapBox recommends prefixes in the style spec.)
2. Cleaned up some linting inconsistencies "if/for(" instead of "if/for (".
3.  Added support for layer titles.
   This support will be necessary come time for making a simple layer list.


__tests__/components/map.test.js
```diff
@@ -44,10 +44,14 @@ describe('Map component', () => {
       "id": "states",
       "source": "states-wms",
     }];
+    const metadata = {
+      'bnd:sources-verison': 0,
```
bartvde: typo: verison instead of version

__tests__/components/map.test.js
```diff
@@ -81,8 +85,11 @@ describe('Map component', () => {
     }];
     const center = [0, 0];
     const zoom = 2;
-    const _sourcesVersion = 0, _layersVersion = 0;
-    const wrapper = shallow(<Map map={{center, zoom, sources, layers, _sourcesVersion, _layersVersion}} />);
+    const metadata = {
+      'bnd:sources-verison': 0,
```
bartvde: typo

__tests__/components/map.test.js
```diff
@@ -103,10 +110,14 @@ describe('Map component', () => {
       "id": "tilejson-layer",
       "source": "tilejson"
     }];
+
+    const metadata = {
+      'bnd:sources-verison': 0,
```
bartvde: typo

__tests__/components/map.test.js
```diff
@@ -126,10 +137,14 @@ describe('Map component', () => {
...
```
bartvde: typo

__tests__/components/map.test.js
```diff
@@ -139,8 +154,10 @@ describe('Map component', () => {
       map: {
         center,
         zoom,
-        _sourcesVersion: 0,
-        _layersVersion: 1,
+        metadata: {
+          'bnd:sources-verison': 0,
```
bartvde: typo

__tests__/components/map.test.js
```diff
@@ -168,8 +185,11 @@ describe('Map component', () => {
     }];
     const center = [0, 0];
     const zoom = 2;
-    const _sourcesVersion = 0, _layersVersion = 0;
-    const wrapper = shallow(<Map map={{center, zoom, sources, layers, _sourcesVersion, _layersVersion}} />);
+    const metadata = {
+      'bnd:sources-verison': 0,
```
bartvde: typo

__tests__/components/map.test.js
```diff
@@ -178,8 +198,10 @@ describe('Map component', () => {
       map: {
         center,
         zoom,
-        _sourcesVersion: 0,
-        _layersVersion: 1,
+        metadata: {
+          'bnd:sources-verison': 0,
```
bartvde: typo

__tests__/components/map.test.js
```diff
@@ -190,8 +212,10 @@ describe('Map component', () => {
       map: {
         center,
         zoom,
-        _sourcesVersion: 0,
-        _layersVersion: 2,
+        metadata: {
+          'bnd:sources-verison': 0,
```
bartvde: typo

__tests__/reducers/map.test.js
```diff
@@ -259,16 +340,18 @@ describe('map reducer', () => {
...
+        'bnd:layers-version': 0,
+      },
       layers: []
     })
   })
 
   it('should handle ADD_FEATURES', () => {
     // since we do not go through ADD_SOURCE we need to set _dataVersion
-    const source = {"_dataVersion": 0, "data": {"type":"FeatureCollection","crs":{"type":"name","properties":{"name":"urn:ogc:def:crs:OGC:1.3:CRS84"}},"features":[{"type":"Feature","properties":{"n":2,"cat":1},"geometry":{"type":"Point","coordinates":[0.5,0.5]}},{"type":"Feature","properties":{"n":3,"cat":2},"geometry":{"type":"Point","coordinates":[0.5,1.5]}}]}};
+    const source = {"metadata" : {"bnd:data-version" : 0}, "data": {"type":"FeatureCollection","crs":{"type":"name","properties":{"name":"urn:ogc:def:crs:OGC:1.3:CRS84"}},"features":[{"type":"Feature","properties":{"n":2,"cat":1},"geometry":{"type":"Point","coordinates":[0.5,0.5]}},{"type":"Feature","properties":{"n":3,"cat":2},"geometry":{"type":"Point","coordinates":[0.5,1.5]}}]}};
```
bartvde: it looks like mapbox only defines metadata at the layer and the root level officially?

bartvde: So maybe we should store it somewhere in the root metadata keyed by the source id?

src/actions/map.js
```diff
@@ -12,10 +12,10 @@ export function setView(center, zoom) {
   }
 }
 
-export function addLayer(layerDef) {
+export function addLayer(layerDef, layerTitle) {
   return {
     type: MAP.ADD_LAYER,
-    layerDef
+    layerDef, layerTitle
```
bartvde: we could also just provide it in the layerDef directly in the metadata or not? Wonder why it was separated, could be a good reason for it ofcourse, but I couldn't think of one

src/reducers/map.js
```diff
@@ -23,71 +39,90 @@ function addLayer(state, action) {
...
+  // add a layer name if specified in the action.
+  if (action.layerTitle) {
+    new_layer.metadata['bnd:title'] = action.layerTitle;
+  }
+
   return Object.assign({}, state, {
-    _layersVersion: state._layersVersion + 1,
     layers: state.layers.concat([new_layer]),
-  });
+  }, incrementVersion(state.metadata, 'bnd:layers-version'));
```
bartvde: maybe we should put this a const on the top of the file?

```const layerVersionKey = 'bnd:layers-version'``` and use that const everywhere else in the code?

src/actions/map.js
theduckylittle: Honestly, I did it as a convenience wrapper for the developer.  The reducer handles converting it to the metadata syntax.  I just through having a layer title would be a common enough situation that easing its use was worthwhile.

bartvde: I'm fine with keeping it this way

theduckylittle:
@bartvde I've updated the code to use constants instead of dropping strings everywhere.  I made some decisions that I would love some feedback about:

1. I created a function called "dataVersionKey" that both the reducer and the map component use.  I  imported the function FROM the reducer into the map module.  I sorta feel like that shouldn't happen and maybe we should cave into creating a "util.js" or similar module for functions that are useful in multiple places.
2. The 'constants.js' file is another one of those 'catch-all' modules.  I did that BEFORE writing the dataVersionKey function.  Maybe 'constants.js' could be moved to a 'constants' object that is apart of a 'util.js' module? Maybe a 'metadata.js' module would be appropriate as both the function and the constants strictly concern metadata-keys.

Concerning the linting: I will get a proper eslint configuration together -- I have been linting manually thus far.

bartvde:
Yeah I think an util module makes sense. I don't have a strong opinion on if constants should be a separate file, I think it is fine as you've done it now.

PR #555, opened 2017-07-07
Ultimate linting

theduckylittle:
This is a bit of a beast that boils down to only really two fixes:
1. Added linting support
2. Fixed all the linting issues.


bartvde:
looks good thanks for handling this

PR #556, opened 2017-07-10
Implement background layers.

theduckylittle:
Adds support for layers with type = background.  Only supports background-color.

### DIFF
--- a/src/components/map.jsx
+++ b/src/components/map.jsx
@@ -7,7 +7,6 @@ import React from 'react';
 import { connect } from 'react-redux';
 
 import getStyleFunction from 'mapbox-to-ol-style';
-import glfun from '@mapbox/mapbox-gl-style-spec/function';
 
 import OlMap from 'ol/map';
 import View from 'ol/view';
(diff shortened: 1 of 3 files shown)


src/components/map.jsx
```diff
@@ -243,6 +254,21 @@ export class Map extends React.Component {
...
+  configureBackground(layer) {
+    // TODO: Right now there is not a good way of using
+    //       the background-opacity attribute, there is no
+    //       DOM-based backgroundOpacity CSS style and applying
+    //       opacity to the element makes it all "fade".
+    if (layer.paint['background-pattern']) {
+      // TODO: We cannot implement the background pattern
+      //       until glyphs/symbology has been implemented.
+    } else {
+      this.mapdiv.style.backgroundColor = layer.paint['background-color'];
```
bartvde: not sure if it's intentional, but this code differs from https://github.com/boundlessgeo/ol-mapbox-style/blob/master/index.js#L183

theduckylittle: I worked to distill down the essential functionality for two main reasons:
1. Supporting changing the background layer by zoom level seemed needlessly complex.
2. The code Andreas wrote is actually a bit wrong because there isn't actual support for a "background-opacity" CSS directive.  If we're going to fully support the spec we're going to need to correct that oversight by either creating an Color to RGBA conversion and/or looking at generating a client-side based PNG which can be used as a background pattern. The PNG-based solution may be optimal because then we could also implement sprite-based backgrounds.

PR #560, opened 2017-07-14
unitTests for ultimateredux branch

CarlyLanglois:
closed earlier PR on this branch to squash commits.
more tests for map actions, reducer, and component. 
Does addfeatures action behave as expected for 'Feature' type on source? 
The test passes as its currently written but doesn't take a properly defined source.


__tests__/reducers/map.test.js
```diff
@@ -389,6 +428,104 @@ describe('map reducer', () => {
     });
   });
 
+  it('should handle ADD_FEATURES to add action features to a source feature', () => {
```
CarlyLanglois: @theduckylittle here's the condition I'm concerned about. Right now this passes but I don't think it does when the source data is more accurate. if any other keys are added onto the source.data object, they are duplicated onto the new source data object, once in their original position on source.data and then again where they should be in source. data.features along with the features added by the action. 
I'm thinking the way that changeData is used in this case vs. the FeatureCollection case may need to be changed.

theduckylittle: You're right, I see how that is probably happening and had to fix another similar bug earlier.  Can you make sure the source is properly defined and let the test fail? Then I'll merge with the known failed test and make a fix.

theduckylittle:
GitHub didn't catch it but I did add a commit that fixes the `changeData` and `removeFeatures` functions that fixes the failing test.

PR #561, opened 2017-07-17
Now with more Popups!

theduckylittle:
1. Popups all inherit from the @boundlessgeo/sdk/components/map/popup
   component.
   The popup component has a main "wrapper" function renderPopup
   that is passed children.  It makes it easier for the sub-classes
   to call renderPopup() and pass in the desired contents.
2. Popups are added using the map's "initialPopups" prop
   or "addPopup" method.
3. Added Demo for the Popups.
4. Added uuid dependency so as not to depend on counters
   for assigning ID's to items.
5. Added support styles for popups.


theduckylittle:
This has also been tested working on touch devices.

examples/popups/app.jsx
```diff
@@ -0,0 +1,199 @@
...
+const store = createStore(combineReducers({
+  map: SdkMapReducer,
+}), window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
+   applyMiddleware(thunkMiddleware));
+
+
+/** A popup for marking features when they
+ *  are selected.
+ */
+class MarkFeaturesPopup extends SdkPopup {
```
wnordmann: Extending Sdkpopup is pretty advanced, would the demo reach a broader audience if we used `SdkPopup.addPopup` and `SdkPopup.renderPopup` over attachign them to `this`

src/components/map.jsx
```diff
@@ -351,6 +360,79 @@ export class Map extends React.Component {
     }
   }
 
+  updatePopups() {
+    const overlays = this.map.getOverlays();
+    const overlays_to_remove = [];
+
+    overlays.forEach((overlay) => {
```
wnordmann: `for(let i=0, ii=overlay.length ...) would be quicker`

src/components/map.jsx
```diff
@@ -351,6 +360,79 @@ export class Map extends React.Component {
...
+      const id = overlay.get('popupId');
+      if (this.popups[id].state.closed !== false) {
+        // mark this for removal
+        overlays_to_remove.push(overlay);
```
wnordmann: why not call removeOverlay here?

examples/popups/app.jsx
theduckylittle: One of the main reason for using this method is that it provides the correct context for arrow functions ( `=>` ) and allows for the correct context for `this`. Otherwise it gets VERY, VERY messy and requires a lot of crossing of the bridges between the React class, the popup contents and OpenLayers.

src/components/map.jsx
```diff
@@ -2,14 +2,18 @@
...
 import PropTypes from 'prop-types';
 import React from 'react';
+import ReactDOM from 'react-dom';
 import { connect } from 'react-redux';
 
 import getStyleFunction from 'mapbox-to-ol-style';
 
 import OlMap from 'ol/map';
 import View from 'ol/view';
+import Overlay from 'ol/overlay';
```
theduckylittle: Realistically sdk/components/map is a "React" version of the OpenLayers map. There are some major structural differences between the old version of SDK and the new one that still support moving to AwesomeMapLibrary/X, specifically:

1. The description of the map state is still done using MapBox GL Styles.
2. The popups, while featuring DOM-esque content are actually still just React Elements. The Map is still solely acting as a *renderer* and is exercising OpenLayers to do it's job.

src/components/map.jsx
theduckylittle: Yes, it's much faster, but it's broken using OpenLayers' `ol.Collection` objects. OpenLayers allows those objects to be mutable in-loop. This makes using normal for-loops problematic.

src/components/map.jsx
theduckylittle: Because if we mutate the operation in-loop, the layer was occasionally receiving undefined objects causing it to error out.

src/components/map.jsx
```diff
@@ -351,6 +360,79 @@ export class Map extends React.Component {
...
+   *  @param {SdkPopup} popup - Instance of SdkPopop or a subclass.
+   *  @param {boolean}  [silent] - When true, do not call updatePopups after adding.
+   *
+   */
+  addPopup(popup, silent = false) {
+    // each popup uses a unique id to relate what is
+    //  in openlayers vs what we have in the react world.
+    const id = uuid.v4();
+
+    const overlay = new Overlay({
```
theduckylittle: I've added autoPan support that will now pan the map to fit the popup.

theduckylittle:
@wnordmann I think I've addressed everything, please check it out!

theduckylittle:
Another update that fixes autoPan and gets rid of the slightly awkward "this" referencing that onClick required in the previous version.

As having a Popup class allows us to abstract the implementation of the Popup (IE how it's rendered on the map) from the definition of the Popup (the class)... I still think having the classes is a good way to go about it.

examples/popups/app.jsx
```diff
@@ -0,0 +1,192 @@
...
+  }));
+
+  // Add 100 random points to the map
+  addPoints('points', 100);
+
+  ReactDOM.render((
+    <SdkMap
+      store={store}
+      initialPopups={[<NullIslandPopup coordinate={[0, 0]} closeable={false} />]}
+      includeFeaturesOnClick
```
wnordmann: includeFeatureOnClick={false}
no reason to leave empty

examples/popups/app.jsx
```diff
@@ -0,0 +1,192 @@
...
+      onClick={(map, xy, features) => {
+        if (features.length === 0) {
+          // no features, :( Let the user know nothing was there.
+          map.addPopup(<SdkPopup coordinate={xy} closeable><i>No features found.</i></SdkPopup>);
+        } else {
+          // Show the super advanced fun popup!
+          map.addPopup(<MarkFeaturesPopup coordinate={xy} features={features} closeable />);
```
wnordmann: closable={false}

src/components/map.jsx
wnordmann: Ok, map.jsx is the OLMap implementation.  I never realized things like Popups would require so much OL interaction

src/components/map.jsx
```diff
@@ -397,6 +410,96 @@ export class Map extends React.Component {
...
+      // create an empty div element for the Popup
+      element: elem,
+      // allow events to pass through, using the default stopevent
+      // container does not allow react to check for events.
+      stopEvent: false,
+      // put the popup into view
+      autoPan: true,
+      autoPanAnimation: {
```
wnordmann: This is UX decisions being made inside the SDK, probably not a big deal, but it's best to keep it dynamic

theduckylittle: It was really the only way to satisfy the "make the popup fit" automatically.  Which, yeah, is a good idea. If we get feedback for a need to not do that we can probably extend the SdkPopup class.  *But* in terms of styling, et al., that's fully controllable via CSS.

examples/popups/app.jsx
theduckylittle: When it is empty, it's implied to be true. ( https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md )

examples/popups/app.jsx
theduckylittle: See above.

PR #563, opened 2017-07-17
add map name action updates map name in state

CarlyLanglois:
Map name in state tree can be changed with a redux action. 
Basic App contains an input component to demonstrate this action. Needs styling if the component will stay in this app.


CarlyLanglois:
yeah good point. I'll change the names. 
I also just realized I didn't think about writing a test for this so I'll do that too.

CarlyLanglois:
@theduckylittle You want me to also just get rid of the name-change demo in the example app?

theduckylittle:
Let's leave it in for now and as we may end up having a follow up discussion on which features to demo where.

theduckylittle:
I tweaked some of the CSS in the basic app so that it wasn't overlapping on narrow screens. Otherwise it all looks good, thanks!

PR #566, opened 2017-07-20
gl layer maxZoom converting to ol layer maxRes

CarlyLanglois:
Helper function in map component converts gl maxZoom val to maxRes val for ol layers 
- only in web mercator projection.

Changes to the basic example app don't perform as expected to re-render the updated layer. A breaking test is included in this PR for a demo but I'll start a new branch for that issue.


CarlyLanglois:
#567 includes these changes

PR #567, opened 2017-07-20
configureLayers handles changes to min/max zoom vals

CarlyLanglois:
this solves part of the issue in #566. 

layer min/max zoom values can be set, converted to ol min/max res values, and then updated both on gl source and ol rendered layer.

### DIFF
--- a/src/components/map.jsx
+++ b/src/components/map.jsx
@@ -184,6 +184,13 @@ function fakeStyle(layer) {
   }, layer.source);
 }
 
+/** Convert a zoom level to resolution,
+    given Spherical Mercator projection.
+ */
+function zoomToResolution(zoomLevel) {
+  return (156543.03392804097 / (2 ** zoomLevel));
+}
+
 
 export class Map extends React.Component {
 
@@ -311,26 +318,36 @@ export class Map extends React.Component {
    */
   configureLayer(sourcesDef, layer) {
     const layer_src = sourcesDef[layer.source];
+    const maxResolution = layer.minZoom ? Math.round(zoomToResolution(layer.minZoom)) : 156544;
+    const minResolution = layer.maxZoom ? zoomToResolution(layer.maxZoom) : 0.0005831682455839253;
 
     switch (layer_src.type) {
       case 'raster':
         return new TileLayer({
           source: this.sources[layer.source],
+          maxResolution,
+          minResolution,
         });
       case 'geojson':
         return new VectorLayer({
           source: this.sources[layer.source],
           style: fakeStyle(layer),
+          maxResolution,
+          minResolution,
         });
       case 'vector':
         return new VectorTileLayer({
           source: this.sources[layer.source],
           style: fakeStyle(layer),
+          maxResolution,
+          minResolution,
         });
       case 'image':
         return new ImageLayer({
           source: this.sources[layer.source],
           opacity: layer.paint ? layer.paint['raster-opacity'] : undefined,
+          maxResolution,
+          minResolution,
         });
       default:
         // pass, let the function return null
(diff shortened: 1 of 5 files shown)


theduckylittle:
This is looking like it's really coming along! Awesome! I've got some line-by-line feedback in the code but there are two notes that apply more globally:
1. It's `minzoom` and `maxzoom` no uppercase letters and the object members are case sensitive in Javascript (see: https://www.mapbox.com/mapbox-gl-js/style-spec/#layer-minzoom ).
2. Make sure to pull in the latest changes from ultimateredux before pushing (it was up to date at the time of the original push but some changes have been subsequently merged in).

CarlyLanglois:
@theduckylittle thanks! I changed the case and pulled ultimate. I don't see the inline comments, though. Did you mean to add some?

theduckylittle:
Ya know, you need to click the big green button BEFORE the comments post. They're now posted inline. Cheers.

src/components/map.jsx
```diff
@@ -311,26 +318,36 @@ export class Map extends React.Component {
    */
   configureLayer(sourcesDef, layer) {
     const layer_src = sourcesDef[layer.source];
+    const maxResolution = layer.minZoom ? Math.round(zoomToResolution(layer.minZoom)) : 156544;
+    const minResolution = layer.maxZoom ? zoomToResolution(layer.maxZoom) : 0.0005831682455839253;
```
CarlyLanglois: In this case we're looking for a resolution value given a zoom value, right? Where does the resolution parameter for `this.map.getView().getZoomForResolution(resolution)` come from?

theduckylittle: Sorry, yes, that should be something more like `this.map.getView().getResolutions()[zoom]`.

CarlyLanglois: getResolutions() will only return values already set for the view. in this case we dont have those resolutions to set yet, no?
https://openlayers.org/en/latest/apidoc/ol.View.html#getResolutions

theduckylittle: Ack, you're correct. I should've traced back the code better. I thought it would populate resolutions internally if none were given, that assumption was incorrect. 

I think we could imitate how setZoom works:
```javascript
_ol_View_.prototype.setZoom = function(zoom) {
  var resolution = this.constrainResolution(
      this.maxResolution_, zoom - this.minZoom_, 0);
  this.setResolution(resolution);
};
```
That gives us a 1:1 between what openlayers is expecting and how it calculates the resolution for a zoom level.

CarlyLanglois: ok i might be missing something. constrainResolution still requires a resolution value so what does it look like to mimic this?

theduckylittle: Okay... really we could write this code...

```javascript
getResolutionForZoom(zoom) {
  // this will have all the projection information set.
  const view = this.map.getView();
 const max_rez = this.map.getMaxResolution();
 return view.contrainResolution(max_rez, zoom - view.getMinZoom(), 0);
}
```

CarlyLanglois:
ok i used the getResolutionForZoom example, thanks.

One thing, though, which might be fine. The ref on the random-points layer on the basic example re-writes min/maxzoom vals. Which is to be expected. But it does mean that if a min or maxzoom is set on the random-points layer when added, that its proper paint object doesn't get rendered.

theduckylittle:
The `ref` functionality is pretty shallow right now.  I wanted to see how it ended up working in practice and we've not gotten that far yet. As it is right now we just do a copy + mixin and then let the new layer live it's own peaceful existence.

src/components/map.jsx
```diff
@@ -362,6 +365,11 @@ export class Map extends React.Component {
   }
 
   configureLayers(sourcesDef, layersDef, layerVersion) {
+    const view = this.map.getView();
+    function getResolutionForZoom(zoom) {
```
theduckylittle: Let's move this function out of the configureLayers definition and into it's own method.  That will make the function more testable. The calls below will need updated to call "this.getResolutionForZoom" but otherwise look good.

CarlyLanglois:
**isolating

src/components/map.jsx
```diff
@@ -362,6 +371,11 @@ export class Map extends React.Component {
   }
 
   configureLayers(sourcesDef, layersDef, layerVersion) {
+    // const view = this.map.getView();
+    // function getResolutionForZoom(zoom) {
+    //   const max_rez = view.getMaxResolution();
+    //   return view.constrainResolution(max_rez, zoom - view.getMinZoom());
+    // }
```
theduckylittle: Please rm the comments!

src/components/map.jsx
bartvde: just catching up here, please note that recently a function got added to openlayers: https://github.com/openlayers/openlayers/pull/6966

PR #569, opened 2017-07-21
Added support for detecting changes in layer styles

theduckylittle:
1. Added support for detecting changes in a layers
   style.
2. Created a demo of changing the paint settings.
3. Small fixes to linting.


src/components/map.jsx
```diff
@@ -191,6 +193,18 @@ function fakeStyle(layer) {
   }, layer.source);
 }
 
+/** Get a layer by it's id
+ */
+function getLayerById(layers, id) {
```
wnordmann: Can this be moved to the util.js

src/components/map.jsx
```diff
@@ -392,6 +423,9 @@ export class Map extends React.Component {
         // move on with the loop if the layer
         //  could not be defined.
         if (layer_def === null) {
+          // WARNING! Continue is used to short circuit
+          //          the rest of this loop.
+          // eslint-disable-next-line no-continue
           continue;
```
wnordmann: could this avoided using an else?

src/components/map.jsx
```diff
@@ -381,7 +412,7 @@ export class Map extends React.Component {
             // layersDef[x] will contain objects which need to be
```
wnordmann: I always like the convention of moving iterator variables down the alphabet, i, j, k, l, ...

examples/index.html
```diff
@@ -26,6 +26,10 @@
         <a href="./clustering/index.html">Clustering Example</a><br>
         Demonstrates clustering points and changing radius.
     </li>
+    <li>
```
wnordmann: I added this, I wanted it to make my review easier.

theduckylittle: :+1:

src/components/map.jsx
theduckylittle: Moved.

src/components/map.jsx
theduckylittle: Renamed

src/components/map.jsx
theduckylittle: Added a "layer !== null" into the if statements below it. Continue is removed.

PR #573, opened 2017-07-25
Source minzoom on source type tileJSON

CarlyLanglois:
Opening this PR to get some input on how to handle minZoom values on an OL tileJSON source. 

I'm looking for some clarification on the relationship btn a tileGrid and tileJSON in this type of course. I expected to call .getTileGrid() and return what appears to be the tileGrid in the source but I'm getting null. Same goes for .getTileJSON(). Is there something that I'm missing in the relationship between the given url and the tileGrid?

### DIFF
--- a/examples/basic/app.jsx
+++ b/examples/basic/app.jsx
@@ -59,6 +59,7 @@ function main() {
   // be an individual Feature or a FeatureCollection.
   store.dispatch(mapActions.addSource('points', {
     type: 'geojson',
+    maxzoom: 6,
     data: {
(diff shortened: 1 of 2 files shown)


theduckylittle:
The min/maxzoom doesn't need to be applied to the source itself.  Not all MapBox styles will have a direct analog in OpenLayers. The best/most-generic way to do this is to change the OL Layer's min/maxresolution settings when either the MB Layer or MB Source min/maxzoom settings change.

PR #574, opened 2017-07-26
Source minzoom (#SDK-599)

CarlyLanglois:
map component configure layers checks for updates to both gl source and gl layer min/max zoom values. Selects the most limiting combination to update ol  layer min/max res values as needed.

updated test to check for zoom conditionals.

#SDK-599


src/components/map.jsx
```diff
@@ -442,16 +439,39 @@ export class Map extends React.Component {
...
       if (layer !== null && layer.id in this.layers) {
         const ol_layer = this.layers[layer.id];
-
-        if (layer.maxzoom) {
-          const minResolution = getResolutionForZoom(this.map, layer.maxzoom);
-          ol_layer.setMinResolution(minResolution);
+        const layer_src = sourcesDef[layer.source];
+
+        // check for min/max zoom changes on sources and layers
+        let minzoom = null;
```
theduckylittle: I think this code is getting kinda long and it would behoove us to make a couple of utility functions (in util.js) for handling these situations where we need to do a comparison but the values are potentially undefined:

```
function getNumericValues(values) {
  let arr = [];
  for (let i = 0, ii = values.length; i < ii; i++) {
    if(!isNaN(values[i])) {
      arr.push(values[i]);
    }
  }
  return arr;
}

function getMin() {
  const numbers = getNumericValues(arguments);
  if(numbers.length === 0) { return undefined; }
  if(numbers.length === 1) { return numbers[0]; }
  return Math.min.apply(this, getNumericValues(arguments));
}

console.log('getNumericValues', getNumericValues([undefined, 2]));

console.log('getMin, no values', getMin());
console.log('getMin, all undefined values', getMin(undefined, undefined));
console.log('getMin, with one defined value', getMin(undefined, 2));
console.log('getMin, all defined values', getMin(3, 4, 5));
```

And if we have a complimentary `getMax` function then this whole code can get substantially shortened to:
```javascript
const minzoom = util.getMax(layer.minzoom, layer_src.minzoom);
if (minzoom) {
  // set the min resolution
}
```

PR #575, opened 2017-07-26
Features can now be drawn on the map!

theduckylittle:
1. Added a special reducer/action set in order to handle
   the drawing functions (as they are not covered by mapbox gl styles).
2. Modified the map component to use drawing.interaction and drawing.sourceName
   props for identifying when a feature is being drawn and on to what.
3. Created an example app with a working demo that includes onFeatureDrawn
   which is the call back provided upon a features drawing completion.
4. Needed to add the canvas module to the dev dependencies.
   By default jsdom/enzyme/jest does not emulate the canvas, they
   depend on the canvas package. This may not be practical to install
   for all dev environments (looking at you, Windows) and may need to
   be shifted to a peerDependency.
5. The map has withRef: true in order to support introspection.

Refs: SDK-500, SDK-535

### DIFF
--- a/__tests__/components/map-drawing.test.jsx
+++ b/__tests__/components/map-drawing.test.jsx
@@ -1,4 +1,4 @@
-/* global it, describe, expect, beforeEach */
+/* global it, describe, expect, beforeEach, spyOn */
 
 /** Specific tests for map interactions.
  *
@@ -13,10 +13,16 @@ import { mount } from 'enzyme';
 
 import { createStore, combineReducers } from 'redux';
 
+import Feature from 'ol/feature';
+import Point from 'ol/geom/point';
+
 import SdkMap from '../../src/components/map';
 import MapReducer from '../../src/reducers/map';
 import DrawingReducer from '../../src/reducers/drawing';
 
+import * as MapActions from '../../src/actions/map';
+
+import { INTERACTIONS } from '../../src/constants';
 import { DRAWING } from '../../src/action-types';
 
 
@@ -57,37 +63,73 @@ describe('Map with drawing reducer', () => {
 if (HAS_CANVAS) {
   describe('Map component with drawing', () => {
     let store = null;
+    let wrapper = null;
 
     beforeEach(() => {
       store = createStore(combineReducers({
         map: MapReducer,
         drawing: DrawingReducer,
       }));
+
+      store.dispatch(MapActions.addSource('test', {
+        type: 'geojson',
+        data: {
+          type: 'FeatureCollection',
+          features: [{
+            type: 'Feature',
+            geometry: {
+              type: 'Point',
+              coordinates: [0, 0],
+            },
+            properties: { },
+          }],
+        },
+      }));
+
+      store.dispatch(MapActions.addLayer({
+        id: 'points',
+        source: 'test',
+        type: 'circle',
+        paint: {
+          'circle-radius': 5,
+        },
+      }));
+
+      wrapper = mount(<SdkMap store={store} />);
     });
 
     it('turns on a drawing tool', () => {
-      const wrapper = mount(<SdkMap store={store} />);
       const sdk_map = wrapper.instance().getWrappedInstance();
       const ol_map = sdk_map.map;
-
       const n_interactions = ol_map.getInteractions().getLength();
 
       store.dispatch({
         type: DRAWING.START,
-        interaction: 'Point',
+        interaction: INTERACTIONS.point,
         sourceName: 'test',
       });
 
       // if the drawing control has been added to the map then
       //  there should be 1 additional interaction.
       expect(ol_map.getInteractions().getLength()).toBe(n_interactions + 1);
+
+      // get the last interaction, which should be the drawing interaction
+      const draw = ol_map.getInteractions().item(n_interactions);
+
+      const dummy_feature = new Feature();
+      dummy_feature.setGeometry(new Point(5, 5));
+      draw.dispatchEvent({
+        type: 'drawend',
+        feature: dummy_feature,
+      });
+
+      store.dispatch(MapActions.setView([-45, -45], 11));
     });
 
     it('turns off a drawing tool', () => {
-      const wrapper = mount(<SdkMap store={store} />);
       store.dispatch({
         type: DRAWING.START,
-        interaction: 'Point',
+        interaction: INTERACTIONS.point,
         sourceName: 'test',
       });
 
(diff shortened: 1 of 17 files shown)


bartvde:
reviewing this one now

examples/drawing/app.jsx
```diff
@@ -0,0 +1,239 @@
...
+      'fill-outline-color': '#f03b20',
+    },
+  }));
+
+  // A counter for the feature IDs
+  let FEATURE_ID = 1;
+
+  // Promises are used here as a way to demonstrate that the
+  // features could be added asynchronously.  This is useful
+  // in cases in which the feature may need validated by the
```
bartvde: validation

examples/drawing/app.jsx
```diff
@@ -0,0 +1,239 @@
...
+        lines: 'LineString',
+        polygons: 'Polygon',
+      };
+
+      // verify the geometry type matches the source.
+      if (feature.geometry.type === geom_types[sourceName]) {
+        // define the new feature with an ID.
+        const new_feature = Object.assign({}, feature, {
+          properties: {
+            id: FEATURE_ID,
```
bartvde: ```id``` should be outside of properties according to the geojson spec or not?

examples/drawing/app.jsx
```diff
@@ -0,0 +1,239 @@
...
+  //  is immutable. The method here is to remove the old feature,
+  //  then immediate create a new one with our changes.
+  const modifyFeature = (map, sourceName, feature) => {
+    store.dispatch(mapActions.removeFeatures(sourceName, ['==', 'id', feature.properties.id]));
+    store.dispatch(mapActions.addFeatures(sourceName, [feature]));
+    // let the user know what happened.
+    error_div.innerHTML = `Feature ${feature.properties.id} modified.`;
+  };
+
+  // Selecting a feature displays it's ID.
```
bartvde: its

src/components/map.jsx
```diff
@@ -264,11 +272,42 @@ export class Map extends React.Component {
...
     // render() from being called.
     return false;
   }
 
+  /** Callback for finished drawings, converts the event's feature
+   *  to GeoJSON and then passes the relevant information on to
+   *  this.props.onFeatureDrawn.
+   */
+  onFeatureEvent(eventType, sourceName, feature) {
+    if (typeof feature !== 'undefined') {
```
bartvde: just wondering why since I'm used to ```feature !== undefined```

src/components/map.jsx
```diff
@@ -676,6 +715,65 @@ export class Map extends React.Component {
     });
   }
 
+  updateInteraction(drawingProps) {
+    // this assumes the interaction is different,
+    //  so the first thing to do is clear out the old interaction
+    if (this.activeInteractions !== null) {
+      for (let i = 0, ii = this.activeInteractions.length; i < ii; i++) {
+        this.map.removeInteraction(this.activeInteractions[i]);
```
bartvde: just a question here, why did you choose removal over setActive?

src/components/map.jsx
```diff
@@ -676,6 +715,65 @@ export class Map extends React.Component {
...
+    }
+
+    if (drawingProps.interaction === INTERACTIONS.modify) {
+      const select = new SelectInteraction({
+        wrapX: false,
+      });
+
+      const modify = new ModifyInteraction({
+        features: select.getFeatures(),
+      });
```
bartvde: is there a way we can avoid recreating the interactions all the time?

src/components/map.jsx
```diff
@@ -676,6 +715,65 @@ export class Map extends React.Component {
...
+
+      modify.on('modifyend', (evt) => {
+        this.onFeatureEvent('modified', drawingProps.sourceName, evt.features.item(0));
+      });
+
+      this.activeInteractions = [select, modify];
+    } else if (drawingProps.interaction === INTERACTIONS.select) {
+      // TODO: Select is typically a single-feature affair but there
+      //       should be support for multiple feature selections in the future.
+      const select = new SelectInteraction({
```
bartvde: our linter is not complaining about a redefinition of select?

src/components/map.jsx
```diff
@@ -676,6 +715,65 @@ export class Map extends React.Component {
...
+          return (layer.getSource() === layer_src);
+        },
+      });
+
+      select.on('select', () => {
+        this.onFeatureEvent('selected', drawingProps.sourcename, select.getFeatures().item(0));
+      });
+
+      this.activeInteractions = [select];
+    } else if (INTERACTIONS.drawing.includes(drawingProps.interaction)) {
```
bartvde: do we need a polyfill for IE?

src/components/map.jsx
```diff
@@ -692,10 +790,17 @@ Map.propTypes = {
...
   }),
+  drawing: PropTypes.shape({
+    interaction: PropTypes.string,
+    sourceName: PropTypes.string,
+  }),
   initialPopups: PropTypes.arrayOf(PropTypes.object),
   setView: PropTypes.func,
   includeFeaturesOnClick: PropTypes.bool,
   onClick: PropTypes.func,
+  onFeatureDrawn: PropTypes.func,
```
bartvde: missing isRequired?

bartvde: ah I see it's in defaultProps

package.json
```diff
@@ -16,6 +16,7 @@
     "babel-preset-es2015": "^6.24.1",
     "babel-preset-react": "^6.24.1",
     "babel-preset-stage-2": "^6.24.1",
+    "canvas": "^1.6.5",
```
theduckylittle: What do you think the best way to do this is? Could we actually mark this as a "peerDependency"? It's only used for testing and I've written in conditionals so that if the canvas module is not there, it will not run those tests.

__tests__/components/map-drawing.test.jsx
```diff
@@ -0,0 +1,223 @@
...
+    it('turns on feature modification', () => {
+      const sdk_map = wrapper.instance().getWrappedInstance();
+      const ol_map = sdk_map.map;
+
+      const interactions = ol_map.getInteractions();
+      const n_interactions = interactions.getLength();
+
+      store.dispatch({
+        type: DRAWING.START,
+        interaction: INTERACTIONS.modify,
```
theduckylittle: I was thinking that instead of writing additional action types we could write more actions to trigger those types of events. That way the reducer and component logic is simpler. (I'll write them into an update commit.)

__tests__/components/map-drawing.test.jsx
```diff
@@ -0,0 +1,223 @@
...
+        sourceName: 'test',
+      });
+
+      // modify differes from select and drawing tools
```
theduckylittle: fixed.

__tests__/components/map-drawing.test.jsx
```diff
@@ -0,0 +1,223 @@
...
+      //  as it requires adding two interactions to the map.
+      expect(interactions.getLength()).toBe(n_interactions + 2);
+
+      // get the modify interaction
+      const select = interactions.item(interactions.getLength() - 2);
+      const modify = interactions.item(interactions.getLength() - 1);
+
+      // configure a spy, when a feature is modifyed onFeatureEvent
```
theduckylittle: fixed.

package.json
bartvde: I think just having a DEVELOPING.md in the repo and documenting it is enough for now?

examples/drawing/app.jsx
theduckylittle: Are you saying "validation" instead of validated? I'd need to rework other parts of the sentence for the tense to make sense.

bartvde: Right, either ```may need to be validated by the server``` or ```may need validation by the server``` IMHO

theduckylittle: ... I changed it to validation.

examples/drawing/app.jsx
theduckylittle: It could be? I was just using it as a unique identifier. It could be changed to "babadook" and be valid for the rest of the test.

src/components/map.jsx
theduckylittle: I just grew accustomed to the typeof version of the test... We can do it either way. Since the parameter is named we don't need to worry about the variable itself being undefined.

src/components/map.jsx
theduckylittle: We could but the creation of them is not that heavy weight of an operation.  They're also only created when they are activated/deactivated and unloading them means there's less to track.

src/components/map.jsx
theduckylittle: No because select is defined in two different code blocks and as a "const", const/let keep their scope local to whichever block they are defined within.

src/components/map.jsx
theduckylittle: I believe that Babel provides the polyfill.

src/components/map.jsx
theduckylittle: :+1:

src/components/map.jsx
theduckylittle: Prevents the interactions from stacking and the memory from the interaction will get freed after setting activeInteractions to null.

__tests__/components/map.test.jsx
```diff
@@ -362,6 +362,40 @@ describe('Map component', () => {
...
+    const store = createStore(combineReducers({
+      map: MapReducer,
+    }));
+    const wrapper = mount(<ConnectedMap store={store} />);
+    const sdk_map = wrapper.instance().getWrappedInstance();
+
+    store.dispatch(MapActions.setView([-45, -45], 11));
+    sdk_map.map.dispatchEvent({
+      type: 'moveend',
+    });
```
theduckylittle: Added a more reasonable test there.

__tests__/components/map.test.jsx
```diff
@@ -362,6 +362,40 @@ describe('Map component', () => {
...
+
+    sdk_map.map.dispatchEvent({
+      type: 'singleclick',
+      coordinate: [0, 0],
+      // this fakes the clicking of the canvas.
+      originalEvent: {
+        // eslint-disable-next-line no-underscore-dangle
+        target: sdk_map.map.getRenderer().canvas_,
+      },
+    });
```
theduckylittle: And also here.

PR #576, opened 2017-07-26
Feature/paint diff

CarlyLanglois:
Paint-Diff Example App allows user to change circle-type points layer fill color, border color, and radius size.

This should just be a first pass, the other paint properties should get tested for updates next.

JIRA SDK-602


theduckylittle:
How much farther did you want to go with this before merging? I think there might be an issue with detecting when a *layout* has changed but that would be a fix to the component.

PR #577, opened 2017-07-31
Improve test coverage

bartvde:
Some of the code didn't have enough coverage, this PR addresses that


bartvde:
This is now ready for review

Coveralls has been re-enabled on PRs so we can track increases / decreases in coverage better from now on

The map component is not fully tested but contains some hard to test code, no easy wins AFAICT

PR #578, opened 2017-07-31
Added hash-based history tracking.

theduckylittle:
The "built-in" component only handles location
but is structured in such a way that it could
be extended to track more items.

To support this query-string handling functions have
been added to util.

The parseQueryString and encodeQueryObject are used
to convert URL encoded query strings to objects and vice-versa.

Refs: SDK-573

### DIFF
--- a/__tests__/components/history.test.jsx
+++ b/__tests__/components/history.test.jsx
@@ -1,0 +1,91 @@
+/* global it, describe, expect, spyOn, beforeEach */
+
+import React from 'react';
+import { shallow, mount } from 'enzyme';
+
+import { createStore, combineReducers } from 'redux';
+import MapReducer from '../../src/reducers/map';
+import { setView } from '../../src/actions/map';
+
+import SdkHashHistory, { HashHistory } from '../../src/components/history';
+
+describe('test the HashHistory component', () => {
+  it('should render the components without error', () => {
+    shallow(<HashHistory />);
+  });
+
+  it('should change the window hash when x,y, or z change', () => {
+    const wrapper = shallow(<HashHistory />);
+
+    // defaults to center=[0,0], zoom=1
+    expect(window.location.hash).toBe('#x=0&y=0&z=1');
+
+    // update x
+    wrapper.setProps({ map: { center: [-1, 0], zoom: 1 } });
+    wrapper.update();
+    expect(window.location.hash).toBe('#x=-1&y=0&z=1');
+
+    // update y
+    wrapper.setProps({ map: { center: [-1, -1], zoom: 1 } });
+    wrapper.update();
+    expect(window.location.hash).toBe('#x=-1&y=-1&z=1');
+
+    // update z
+    wrapper.setProps({ map: { center: [-1, -1], zoom: 11 } });
+    wrapper.update();
+    expect(window.location.hash).toBe('#x=-1&y=-1&z=11');
+  });
+
+  it('should update the view when x,y,z changes', () => {
+    const props = {
+      setView: () => { },
+    };
+    spyOn(props, 'setView');
+    mount(<HashHistory {...props} />);
+
+    // simulate a hash change event.
+    window.location.hash = '#x=200&y=200&z=22';
+    window.dispatchEvent(new Event('hashchange'));
+
+    // setView is the function property to handle
+    //  changing the view when the hash has changed.
+    expect(props.setView).toHaveBeenCalled();
+  });
+
+  it('should fail gracefully when there is an invalid center value', () => {
+    mount(<HashHistory />);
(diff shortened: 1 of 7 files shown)


src/util.js
```diff
@@ -52,3 +52,41 @@ export function getMax(...args) {
...
   return Math.max.apply(this, getNumericValues(args));
 }
+
+/** Parse an arbitrary string as if it were a URL.
+ *
+ *  @param queryString {String} - The query string to parse.
+ *
+ * @returns An object with the key-value-pairs.
+ */
+export function parseQueryString(queryString) {
```
bartvde: just wondering why we roll our own over well-tested small libs like: https://www.npmjs.com/package/url-parse ?

bartvde: btw, I used that one before in sdk so I'm biased :)

__tests__/components/history.test.jsx
```diff
@@ -0,0 +1,91 @@
...
+describe('test the HashHistory component', () => {
+  it('should render the components without error', () => {
+    shallow(<HashHistory />);
+  });
+
+  it('should change the window hash when x,y, or z change', () => {
+    const wrapper = shallow(<HashHistory />);
+
+    // defaults to center=[0,0], zoom=1
+    expect(window.location.hash).toBe('#x=0&y=0&z=1');
```
bartvde: just note that we will be breaking all previous urls by changing the format, old format was:

```#map=9783.93962050256/-16839563.6/8850169.51/0```

I had issues with the zoom level in openlayers back then (can't recall the exact details unfortunately), hence I went with resolution. Also rotation is in the hash.

Comment (bartvde) [inline __tests__/components/history.test.jsx]: Ah maybe this was before Andreas fixed: https://github.com/openlayers/openlayers/pull/5674 i.e. my issues with zoom

File: __tests__/root/util.test.js [hunk 3]
```diff
@@ -41,4 +41,16 @@ describe('util', () => {
     expect(util.jsonEquals(paint, new_paint)).toEqual(false);
     expect(util.jsonEquals(paint, same_paint)).toEqual(true);
   });
+
+  it('parses aand deodes a query string', () => {
```
bartvde: decodes

__tests__/components/history.test.jsx
bartvde: Btw, I'm not opposed to change the format of the hash for good reasons

src/components/history.js
```diff
@@ -0,0 +1,152 @@
...
+    // without this, render() will overwrite what is in the URL.
+    this.onHashChange();
+  }
+
+  /** After the component has mounted,
+   *  add a listener for the history changes.
+   */
+  componentDidMount() {
+    // addEventListener is supported in IE9+
+    window.addEventListener('hashchange', () => {
```
bartvde: what about ```popstate``` and using the history API of the browser? So people can navigate easier?

src/components/history.js
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  /** Create an object which will be encoded in the hash
+   *
+   *  This object should represent a limited amount of state
+   *  which is appropriate for hash-encoding.
+   *
+   */
+  getStateObject() {
+    return Object.assign(
```
bartvde: why do we need to copy the object here?

src/components/history.js
```diff
@@ -0,0 +1,152 @@
...
+
+  /** Parse the location from the hash.
+   *
+   *  @param parsedHash The object resulting from parseQueryString of the hash.
+   *
+   */
+  parseLocationHash(parsedHash) {
+    // create copies of the current center and zoom
+    const center = this.props.map.center.slice();
+    let zoom = 0 + this.props.map.zoom;
```
bartvde: is this really needed for numbers? I thought they got always copied?

src/components/history.js
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  /** Parse the location from the hash.
+   *
+   *  @param parsedHash The object resulting from parseQueryString of the hash.
+   *
+   */
+  parseLocationHash(parsedHash) {
+    // create copies of the current center and zoom
+    const center = this.props.map.center.slice();
```
bartvde: instead of slicing and changing below, can't we just create new array?

__tests__/root/util.test.js
theduckylittle: fixed.

src/components/history.js
theduckylittle: Using hash will also let the users navigate using the back/forward buttons.

src/components/history.js
theduckylittle: I did this so that sub-classes can override the getStateObject function and add additional parameters to be URL tracked.

PR #580, opened 2017-08-01
Add export image functionality

bartvde:
SDK-511


examples/export-image/app.jsx
```diff
@@ -0,0 +1,57 @@
...
+  map: SdkMapReducer,
+  print: SdkPrintReducer,
+}), window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
+   applyMiddleware(thunkMiddleware));
+
+function main() {
+  const url = 'https://raw.githubusercontent.com/boundlessgeo/ol-mapbox-style/master/example/data/wms.json';
+  store.dispatch(mapActions.setContext({ url }));
+
+  const exportMapImage = (blob) => {
```
wnordmann: A comment here about how this is the function called by the onExportImage prop would be useful

examples/export-image/app.jsx
```diff
@@ -0,0 +1,57 @@
...
+    saveAs(blob, 'map.png');
+    store.dispatch(printActions.receiveMapImage());
+  };
+
+  // place the map on the page.
+  ReactDOM.render(
```
wnordmann: I think the export would be more interesting if there some some data included, even the random points would demo that layers are included in the export

src/components/map.jsx
```diff
@@ -278,6 +278,11 @@ export class Map extends React.Component {
       this.updateInteraction(nextProps.drawing);
     }
 
+    if (nextProps.print && nextProps.print.exportImage) {
```
wnordmann: a comment here about using the canvas api would be helpful

examples/export-image/app.jsx
bartvde: good point, will add

examples/export-image/app.jsx
bartvde: there is the WMS USA states overlay, but it's currently not visible due to the center issue that duck just fixed, so should resolve itself

src/components/map.jsx
bartvde: will add

examples/export-image/app.jsx
bartvde: I've rebased this branch so the center is now ok and shows the USA states layer for printing

bartvde:
@wnordmann I've addressed your review with: https://github.com/boundlessgeo/sdk/pull/580/commits/82fb8d3e8458a4f6b43ae58cb9677a2819548313

@duck I will await your review

bartvde:
@duck look at the example, once the image is received the toggle is flipped. So pressing the button again would work and result in a new image.

bartvde:
@duck I didn't really like that the example code is responsible for flipping the switch though, but couldn't think of anything better, maybe you have other ideas for that

bartvde:
@duck I just rebased this, let me know if you need any changes after I replied to your comment

theduckylittle:
@bartvde  -- note my handle on here is @theduckylittle ... so someone else is probably really curious as to why they're getting notifications.

bartvde:
ha, good catch ;-)

PR #582, opened 2017-08-03
Added new legend component and example for it.

theduckylittle:
The new legend component uses a combination of the state
and metadata in order to render one of four differnet types of
legends:

1. A WMS legend.  When pointing to a Raster source which uses
   WMS for tiles, the legend component will use the URL to
   generate a set of images for each layer in the source.
2. html-type legends can contain HTML in the "bnd:legend-content" metadata.
3. image-type legends can refer to an image URL in the
   "bnd:legend-content" which will then include that image for the legend.
4. href-type legends specify a URL in "bnd:legend-content" which
   will be fetched and then rendered in the div.
5. Added tests


__tests__/components/legend.test.jsx
```diff
@@ -0,0 +1,149 @@
...
+      done();
+    }, 300);
+  });
+
+  it('should test the null state', () => {
+    mount(<SdkLegend layerId="null-test" store={store} />);
+  });
+
+  it('should fail miserably', () => {
+    mount(<SdkLegend layerId="does-not-exist" store={store} />);
```
bartvde: but this does not fail right?

examples/legends/app.jsx
```diff
@@ -0,0 +1,165 @@
+/** Very basic SDK application example.
+ *
+ *  Contains a Map and demonstrates some of the dynamics of
+ *  using the store.
+ *
```
bartvde: copy/paste header? outdated

examples/legends/app.jsx
```diff
@@ -0,0 +1,165 @@
...
+
+// This will have webpack include all of the SDK styles.
+import '@boundlessgeo/sdk/stylesheet/sdk.css';
+
+/* eslint-disable no-underscore-dangle */
+const store = createStore(combineReducers({
+  map: SdkMapReducer,
+}), window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
+   applyMiddleware(thunkMiddleware));
+window.store = store;
```
bartvde: was this used for testing?

examples/legends/app.jsx
```diff
@@ -0,0 +1,165 @@
...
+
+function main() {
+  // Start with a reasonable global view of the map.
+  store.dispatch(mapActions.setView([-90, 0], 2));
+
+  store.dispatch(mapActions.setMapName('Basic Map Example'));
```
bartvde: Legend Example?

examples/legends/app.jsx
```diff
@@ -0,0 +1,165 @@
...
+
+  store.dispatch(mapActions.addLayer({
+    id: 'states',
+    source: 'states',
+  }));
+
+   // place the map on the page.
+  ReactDOM.render(<SdkMap store={store} />, document.getElementById('map'));
+
+  // add some buttons to demo some actions.
```
bartvde: there are no buttons here ;-)

src/components/legend.jsx
```diff
@@ -0,0 +1,212 @@
...
+import fetch from 'isomorphic-fetch';
+import PropTypes from 'prop-types';
+import React from 'react';
+import { connect } from 'react-redux';
+
+import { parseQueryString, encodeQueryObject } from '../util';
+
+
+/** Return a div that is asynchronously populated
+ *  with the content from the parmaeter href.
```
bartvde: parameter

src/components/legend.jsx
```diff
@@ -0,0 +1,212 @@
...
+function getRemoteLegend(href) {
+  let ref = null;
+  const div = (<div ref={(me) => { ref = me; }} />);
+
+  // kick off the href fun!
+  fetch(href).then(response => response.text())
+  .then((html) => {
+    // This is equivalent to dangerouslySetInnerHTML
+    if (ref !== null) {
+      ref.innerHTML = html;
```
bartvde: do you see any XSS vulnerabilty possibilities here? Or we should be safe?

src/components/legend.jsx
```diff
@@ -0,0 +1,212 @@
...
+ *  additional information.
+ *  "bnd:legend-type" : "image", "bnd:legend-content" would provide
+ *   the src attribute for an <img>
+ *  "bnd:legend-type" : "html", "bnd:legend-content" would provide
+ *   the html content for a <div>
+ *  "bnd:legend-type" : "href", "bnd:legend-content" would provide
+ *   the URL for html content.
+ */
+export function getVectorLegend(layer) {
+  const metadata = layer.metadata ? layer.metadata: {};
```
bartvde: if there is no metadata we can bail out earlier right?

src/components/legend.jsx
```diff
@@ -0,0 +1,212 @@
...
+  const content = metadata['bnd:legend-content'];
+
+  switch (metadata['bnd:legend-type']) {
+    case 'image':
+      return (<img alt="vector legend" src={content} />);
```
bartvde: does this mean we need i18n soon now we are adding our first user interface strings :) ?

src/components/legend.jsx
```diff
@@ -0,0 +1,212 @@
...
+      return getRemoteLegend(content);
+    default:
+      // no legend provided.
+      return null;
+  }
+}
+
+/** Get the legend for a raster-type layer.
+ *
+ *  Attempts to detect a WMS-type source and use GetFeatureInfo,
```
bartvde: I guess you mean GetLegendGraphic

src/components/legend.jsx
```diff
@@ -0,0 +1,212 @@
...
+      //  sometimes lower cased, and sometimes (evilly so) mixed case.
+      const wms_keys = Object.keys(wms_params);
+      for (let i = 0, ii = wms_keys.length; i < ii; i++) {
+        const key = wms_keys[i];
+        const uc_key = key.toUpperCase();
+        wms_params[uc_key] = wms_params[key];
+      }
+
+      // get the WMS servers URL.
+      const url = tile_url.split('?')[0];
```
bartvde: I know this is inexpensive but we can use split once, we are now doing it twice

src/components/legend.jsx
```diff
@@ -0,0 +1,212 @@
...
+        });
+        const src = `${url}?${encodeQueryObject(params)}`;
+        images.push((<img alt="Legend" key={layers[i]} className="sdk-legend-image" src={src} />));
+      }
+
+      return images;
+    }
+  }
+
+  return getVectorLegend(layer);
```
bartvde: I'd probably call this method getLegend since it has common parts for raster and vector

__tests__/components/legend.test.jsx
```diff
@@ -0,0 +1,149 @@
...
+      },
+    });
+  });
+
+  afterEach(() => {
+    nock.cleanAll();
+  });
+
+  it('should render the legend without error', () => {
+    shallow(<SdkLegend layerId="wms" store={store} />);
```
theduckylittle: I updated this to use a real id, and switched from shallow to mount. Shallow wasn't actually rendering the legend.

__tests__/components/legend.test.jsx
theduckylittle: This is now testing for a legitimately empty legend.

examples/legends/app.jsx
theduckylittle: Yes, removed. Some folks use some fancy extension -- I use the dev console. I'm thinking I should write a eslint rule to prevent me from committing that specific line of code.

examples/legends/app.jsx
theduckylittle: Unneeded line of code. Removed.

examples/legends/app.jsx
theduckylittle: s/buttons/legends/ :wq

src/components/legend.jsx
theduckylittle: :+1:

src/components/legend.jsx
theduckylittle: Yep, good catch.

src/components/legend.jsx
theduckylittle: This has a potential for XSS issues. However, I have a few answers for that:
1. I think we're going to need it. Especially if we implement AGOL legends which basically require making a fetch call to get HTML.
2. We could choose a "template" engine that would make this safer.  Like Mark.up or something similar.  I'm open minded.

src/components/legend.jsx
theduckylittle: Swear to god, I wrote that comment while I was chatting about fixing GeoJson stuff in MapServer and mixed my channels.

src/components/legend.jsx
theduckylittle: Maybe? I'd like to avoid needing i18n support but without alt-tags React will shout a bout the image not being "safe." My thoughts on this:
1. We put the fixed alt-string into props (that way they can be subbed out).
2. Switch the alts to just the layer-name no matter what.  I could also see us adding a "bnd:layer-label" or even "bnd:layer-legend-label" metadata that would give us more options.

src/components/legend.jsx
theduckylittle: :+1:

src/components/legend.jsx
bartvde: both options work for me, no real preference

theduckylittle: *For now*, I've changed it to the layer name.  I suspect we'll need to address that change in the future but I'd like feedback for how people want to label the legends.

bartvde:
92% ;-) nice

PR #583, opened 2017-08-03
Layer list

wnordmann:
SDK-619
This PR is a little bigger than expected because my branch is ahead from the sampleApp feature table pr.

There was an unexpected bug in reducers/map.js about ordering the layers.  Not sure this is the best solution.  I'd like to write tests after getting feedback on the new reorederLayers function


theduckylittle:
@wnordmann  was this based on your sampleApp branch and not ultimateredux???

theduckylittle:
@wnordmann -- Why do we need the new reorderLayer function? `placeLayer()` and `orderLayer()` provide the reducer side of this functionality...

PR #586, opened 2017-08-07
Add more complete WMS example

bartvde:
SDK-515

### DIFF
--- a/examples/wms/app.jsx
+++ b/examples/wms/app.jsx
@@ -98,8 +98,9 @@ function main() {
       const view = map.map.getView();
       const layers = state.map.layers;
       for (let i = 0, ii = layers.length; i < ii; ++i) {
-        if (layers[i].metadata['bnd:queryable']) {
-          const url = map.sources[layers[i].source].getGetFeatureInfoUrl(
+        const layer = layers[i];
+        if (layer.metadata['bnd:queryable'] && (!layer.layout || (layer.layout.visibility && layer.layout.visibility !== 'none'))) {
+          const url = map.sources[layer.source].getGetFeatureInfoUrl(
             xy.xy, view.getResolution(), view.getProjection(), {
               INFO_FORMAT: 'application/json',
             },
--- a/examples/wms/app.jsx
+++ b/examples/wms/app.jsx
@@ -9,6 +9,7 @@ import thunkMiddleware from 'redux-thunk';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import WMSCapabilitiesFormat from 'ol/format/wmscapabilities';
+import TileWMSSource from 'ol/source/tilewms';
 
 import SdkMap from '@boundlessgeo/sdk/components/map';
 import SdkMapReducer from '@boundlessgeo/sdk/reducers/map';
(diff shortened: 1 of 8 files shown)


examples/wms/app.jsx
```diff
@@ -118,27 +55,81 @@ function main() {
...
-  ReactDOM.render(<SdkMap store={store} />, document.getElementById('map'));
+  ReactDOM.render(<SdkMap
+    onClick={(map, xy) => {
+      // show a popup containing WMS GetFeatureInfo.
+      const state = store.getState();
+      const view = map.map.getView();
+      const layers = state.map.layers;
+      for (let i = 0, ii = layers.length; i < ii; ++i) {
+        const layer = layers[i];
+        if (layer.metadata['bnd:queryable'] && (!layer.layout || (layer.layout.visibility && layer.layout.visibility !== 'none'))) {
```
theduckylittle: This should check to ensure that the layer is attached to a WMS-type source.  Queryable could also be set to "true" for a WFS source which would not use the getFeatureInfoUrl request.

examples/wms/app.jsx
```diff
@@ -118,27 +55,81 @@ function main() {
...
+      }
+        layers={root.Layer}
+      />, document.getElementById('add-wms'));
+    });
+  };
 
   // place the map on the page.
-  ReactDOM.render(<SdkMap store={store} />, document.getElementById('map'));
+  ReactDOM.render(<SdkMap
+    onClick={(map, xy) => {
```
theduckylittle: I wonder if we should modify onClick to work with promises. I think I had originally done that and then bailed due to some scoping issues that I was fighting.  So when "includeFeatures" was set to true, onClick could use `(map, xy, featuresPromise)` as the parameters for the callback.

Then the code body would be something like:
```javascript
onClick=(map, xy, featuresPromise) {
  featuresPromise.then((features) => {
   // add popups from here.
 });
}
```

This has a few advantages:
1. It ensures that no matter the type of source queried, the code would always look the same to the end user.
2. It would merge features easily with vector sources.

theduckylittle:
I accidentally put "request changes" when I should've more appropriate said, "comment." My thoughts on the code are inline and certainly debatable and meant to start discussion.

examples/wms/app.jsx
bartvde: Check added

examples/wms/app.jsx
bartvde: It might make sense indeed, one question, would you suggest that the retrieval of WMS GetFeatureInfo would be moved from the example code into the Map component? Or that step is not needed?

bartvde: I'm probably missing something here, but we could simply do:

```this.props.onClick(this, coordinate, new Promise((resolve) => { resolve(click_features); })/*click_features*/);``` 

?

bartvde:
@theduckylittle no problem, let's discuss the way forward when you get online today

bartvde:
I need to add tests to increase the coverage after the last change

bartvde:
@theduckylittle changes have been made to use promises, can you have another look please?

theduckylittle:
Looks great!

PR #587, opened 2017-08-07
Additional placeLayer/orderLayer tests.

theduckylittle:
This includes a more fully thought out version
of the tests for re-ordering layers. It features four
layers and tests placing one in the middle, end, and beginning
of the layer list.

### DIFF
--- a/__tests__/reducers/map.test.js
+++ b/__tests__/reducers/map.test.js
@@ -1,9 +1,11 @@
-/* global it, describe, expect */
+/* global it, describe, expect, beforeEach */
 
 import deepFreeze from 'deep-freeze';
 
 import reducer from '../../src/reducers/map';
 import { MAP } from '../../src/action-types';
+import * as MapActions from '../../src/actions/map';
+
 
 describe('map reducer', () => {
   it('should return the initial state', () => {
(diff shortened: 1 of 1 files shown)


__tests__/reducers/map.test.js
```diff
@@ -1002,3 +1004,63 @@ describe('map reducer', () => {
...
+    runOrderTest('layer_a', undefined, [layer_b, layer_c, layer_d, layer_a]);
+  });
+
+  it('moves a layer to the beginning', () => {
+    runOrderTest('layer_d', 'layer_a', [layer_d, layer_a, layer_b, layer_c]);
+  });
+
+  it('moves a layer to the middle', () => {
+    runOrderTest('layer_d', 'layer_b', [layer_a, layer_d, layer_b, layer_c]);
+  });
```
wnordmann:
```
  it('moves a middle top layer up to the top', () => {
    runOrderTest('layer_c', 'layer_d', [layer_a, layer_b, layer_d, layer_c]);
  });
```
this test still fails

theduckylittle: This doesn't *swap* layers. To move a layer to the top you'd pass null (the end of the list).  So if you're trying to push a `layer_c` to the top, it would be...
```javascript
    runOrderTest('layer_c',  undefined, [layer_a, layer_b, layer_d, layer_c]);
```

wnordmann: what about ```it('moves a layer in the the middle up', () => {
    runOrderTest('layer_b', 'layer_c', [layer_a, layer_c, layer_b, layer_d]);
  });```
?

theduckylittle: 'layer_b', 'layer_c', would be a non-op since 'layer_b' is already "below" 'layer_c'. To move 'layer_b' at top 'layer_c' you would need to do 'layer_c', 'layer_b', or 'layer_b', 'layer_d'.

wnordmann: Is there a reason not to include a reorder function?  That would easily allow the swapping of layers.  The common use case would be a drag and drop layer list where a user would arbitrarily move layers around.

theduckylittle: There are a few reasons:
1. Leaving it out means less API. That's less to test and less code overall to maintain.
2. This form of the function is actually similar to how the MapBox JS library handles the same problem. 
3. All the information should be there already.  In the case of "dropping" a layer, you're not actually arbitrarily moving the layers around.  There's a "dropped on" layer that is either shifted before or after the "dragged" layer.  If the application behaves similar to the API (where the dropped on layer goes one ahead of the dragged layer) then you'd need to make a call similar to this: `orderLayers(dragged.id, dropped.id)`.

PR #588, opened 2017-08-07
layout diff updates sprite icon

CarlyLanglois:
layout is checked for updates, causing re-render

the map component test labelled WORKING TEST does not currently create an ol Style object when an 'icon-image' property is supplied to layout. I think I'm still missing some criteria in the test to properly pass in the sprite data but not quite getting it.


bartvde:
@CarlyLanglois @theduckylittle any idea why in the current code it reads map.sprites instead of map.sprite as in the mapbox style spec?

AFAICT the sprite configuration is async so it will not happen before you call the style function in your test

CarlyLanglois:
JIRA #SDK-620

PR #590, opened 2017-08-08
Increase map component coverage

bartvde:
This was the only component with little coverage left


bartvde:
Somehow overlay.getElement() resulted in warnings for unmount in the test framework, storing the element directly did work however.

PR #593, opened 2017-08-09
Adding in a simple map prop from zoom slider

wnordmann:
Zoom slider prop added to map with some tests

Added zoomslider to basic example

Adding tests and fixing lint

### DIFF
--- a/examples/basic/app.jsx
+++ b/examples/basic/app.jsx
@@ -256,7 +256,7 @@ function main() {
   };
 
   // place the map on the page.
-  ReactDOM.render(<SdkMap store={store} zoomSlider />, document.getElementById('map'));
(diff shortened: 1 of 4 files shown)


examples/basic/app.jsx
```diff
@@ -256,7 +256,7 @@ function main() {
   };
 
   // place the map on the page.
-  ReactDOM.render(<SdkMap store={store} />, document.getElementById('map'));
+  ReactDOM.render(<SdkMap store={store} zoomSlider />, document.getElementById('map'));
```
theduckylittle: Can we make this more self-documenting like "showZoomSlider" or includeZoomSlider?

theduckylittle:
Looks good, just one comment about makign the prop more self documenting!

PR #594, opened 2017-08-09
Rotating Map Example App JIRA SDK-605

CarlyLanglois:
map bearing value in degrees is stored in state and converted to rotation radians for ol def.
example app rotates map clockwise and counterclockwise.


examples/rotating/app.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import SdkMap from '@boundlessgeo/sdk/components/map';
+import SdkMapReducer from '@boundlessgeo/sdk/reducers/map';
+import * as mapActions from '@boundlessgeo/sdk/actions/map';
+
+// This will have webpack include all of the SDK styles.
+import '@boundlessgeo/sdk/stylesheet/sdk.css';
```
bartvde: this should now be scss?

PR #596, opened 2017-08-10
Fix for single-feature regression.

theduckylittle:
After the latest set of merges, I found that a single feature
was not rendering on the map. The check for "source.data.features"
was not properly testing for the case where there could be a single
feature stored in "data".  I've updated teh check and it appears to
be working fine now.


bartvde:
any chance for a unit test?

bartvde:
btw looks that according to the spec data can also be url, but we don't support that as yet right?

https://www.mapbox.com/mapbox-gl-js/style-spec/#sources-geojson

theduckylittle:
No we do not. But we should draft that ticket and include it in the next sprint.

theduckylittle:
I'll work on a unit test for this today. I suspect the test will require a lot of setup and take a while to write.

bartvde:
I will create a JIRA

src/components/map.jsx
```diff
@@ -127,7 +127,7 @@ function updateGeojsonSource(olSource, glSource, opt_mapProjection = 'EPSG:3857'
 
   let features;
 
-  if (glSource.data.features) {
+  if (glSource.data.type === 'Feature' || glSource.data.features) {
```
bartvde: wouldn't a check for data be sufficient here?

bartvde: i.e. ```if(glSource.data)```

theduckylittle: I would think so but I want to write the tests now because I'm not sure how fragile the parser can be when it is given bad input data.

PR #598, opened 2017-08-10
patch to read features from source data w/o addfeature action

CarlyLanglois:
basic app and sprites app are missing features on initial load w/o an addFeatures action specified in the apps, even when the feature data is populated in the addSource action.
this should restore the initial rendering of geojson layers w/ features.


CarlyLanglois:
closing b/c this is a dupe of #596

PR #601, opened 2017-08-11
starting geolocation app

CarlyLanglois:
JIRA SDK-575  

point added and zoomed to when geolocation button is clicked

adding remove feature option to toggle location

locating div in place when location info is being retrieved

### DIFF
--- a/examples/geolocation/app.jsx
+++ b/examples/geolocation/app.jsx
@@ -76,26 +76,26 @@ function main() {
     },
   }));
 
-  function success(position) {
-    document.getElementById('status').innerHTML = '';
-    const latitude = position.coords.latitude;
-    const longitude = position.coords.longitude;
-    store.dispatch(mapActions.addFeatures('points', [{
-      type: 'Feature',
-      properties: {
-        title: 'User Location',
-      },
-      geometry: {
-        type: 'Point',
-        coordinates: [longitude, latitude],
-      },
-    }]));
-    store.dispatch(mapActions.setView([longitude, latitude], 18));
-  }
+  // function success(position) {
+  //   document.getElementById('status').innerHTML = '';
+  //   const latitude = position.coords.latitude;
+  //   const longitude = position.coords.longitude;
+  //   store.dispatch(mapActions.addFeatures('points', [{
+  //     type: 'Feature',
+  //     properties: {
+  //       title: 'User Location',
+  //     },
+  //     geometry: {
+  //       type: 'Point',
+  //       coordinates: [longitude, latitude],
+  //     },
+  //   }]));
+  //   store.dispatch(mapActions.setView([longitude, latitude], 18));
+  // }
 
-  function error() {
-    document.getElementById('status').innerHTML = 'Unable to retrieve your location';
-  }
+  // function error() {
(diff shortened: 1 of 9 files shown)


examples/geolocation/app.jsx
```diff
@@ -0,0 +1,126 @@
...
+    type: 'circle',
+    paint: {
+      'circle-radius': 13,
+      'circle-color': '#149BE9',
+      'circle-stroke-color': '#f149BE9',
+    },
+  }));
+
+  function success(position) {
+    ReactDOM.unmountComponentAtNode(document.getElementById('loading'));
```
theduckylittle: With the way this is being used, it's easier to just use `innerHTML`.  That'll be substantially less resource intensive and markedly less confusing.

examples/geolocation/app.jsx
```diff
@@ -0,0 +1,126 @@
...
+      geometry: {
+        type: 'Point',
+        coordinates: [longitude, latitude],
+      },
+    }]));
+    store.dispatch(mapActions.setView([longitude, latitude], 18));
+  }
+
+  function error() {
+    ReactDOM.render(<p>Unable to retrieve your location</p>, document.getElementById('controls'));
```
theduckylittle: Same here, we really don't need to be rendering/unmounting DIV content.

theduckylittle:
I think this is a reasonable start but I also think we could flush this out more fully:

1. Create a new TrackPosition component. ('track-position.jsx' in the sample application directory). Track positoin should have the following props:
  * refresjInterval : number of ms between location updates.
  * targetSource: The source-name which TrackPosition can add/update/remove points from. This should also be able to handle when targetSource is undefined.
  * showLocation: When true, populate `current_location` with a pretty version of the users coordinates.
2. The TrackPosition component is connected to the store and renders a button...
```javascript
<span>
  <button>{ loading_icon } { status }</button> { current_location }
</span>
```
The status changes between: 'Start Tracking' and 'Stop Tracking'.  The loading icon shows something (maybe a spinny-gif) whenever the locator is actually running. current_location is a pretty string that shows the users location when `showLocation` is true.

3. Instead of the TrackPosition component doing a one-time location it will poll for the location every `refreshInterval` milliseconds.  When the location updates, it will re-center the map to that new location (dispatch a setView action with a new center) and move the "point" it created.

examples/geolocation/app.jsx
```diff
@@ -0,0 +1,126 @@
...
+    id: 'background',
+    type: 'background',
+    paint: {
+      'background-color': '#eee',
+    },
+  }));
+
+  // Show null island as a layer.
+  store.dispatch(mapActions.addLayer({
+    id: 'null-island',
```
theduckylittle: Let's rename this (and update the comment above) to make it more meaningful for this application as `null-island` doesn't really apply here.

bartvde:
looks like the commit history and the diff for this is not clean?

CarlyLanglois:
thanks Bart! Yes, this isnt ready for a merge. I had a feeling this would happen but I wanted to push to keep some changes safe before I pulled down new code. I'll close this pr and open a new one when it's ready to go

PR #603, opened 2017-08-14
Handle undefined center, zoom, bearing in map component

bartvde:
SDK-630 and SDK-633


bartvde:
I think it's still useful to have 0 for the default bearing in the state. OpenLayers uses the same default.

When using setContext, basically source and layer revisions would be reset to 0 again, they would start afresh.

Thanks for the review.

PR #609, opened 2017-08-15
Update branch list in Travis config

bartvde:
This updates our Travis config with the correct set of branches


bartvde:
config only so merging

PR #610, opened 2017-08-16
Make sure we pass on map projection to clustering logic

bartvde:
SDK-636


bartvde:
This is now ready for review

PR #612, opened 2017-08-17
Fixed cluster reprojection bug.

theduckylittle:
When updating sources, clusters were not properly
reprojection on update.


theduckylittle:
To recreate on master:
* open the clustering app
* click the cluster checkbox.
* click it again.

It won't work well. Everything gets grouped into the center.  This fix ensures the projection is handled properly.

bartvde:
see #610 which does the same but has tests

theduckylittle:
Closing as a dup.

PR #613, opened 2017-08-17
 Grouping GL Layers into OL Layers for performance optimizations

theduckylittle:
1. Modified the SDK GL->OL mapping to do grouping.
2. Modified the tests (mostly) to reflect this.
3. This includes PR #612 as I found it during the testing process.
4. This upgrades the versions of the @ahocevar library dependencies as they use a new "apply" style function which is different from previous.
5. Added an "mb-bright" demo that will use the example mapbox bright stylesheet for maps.

Thing that *are* working:
1. Almost everything appears to work on the surface.
2. Changing map.sprite does not update the underlaying layers right now.
3. The MVT performance is now really good, no more multi-layer stuff.


examples/mb-bright/app.jsx
```diff
@@ -0,0 +1,62 @@
+/* global saveAs */
+/** Demo of using the drawing, modify, and select interactions.
```
bartvde: outdated comment

examples/mb-bright/app.jsx
```diff
@@ -0,0 +1,62 @@
...
+import SdkMap from '@boundlessgeo/sdk/components/map';
+import SdkMapReducer from '@boundlessgeo/sdk/reducers/map';
+import SdkPrintReducer from '@boundlessgeo/sdk/reducers/print';
+import * as mapActions from '@boundlessgeo/sdk/actions/map';
+import * as printActions from '@boundlessgeo/sdk/actions/print';
+
+// This will have webpack include all of the SDK styles.
+import '@boundlessgeo/sdk/stylesheet/sdk.scss';
+
+import CONFIG from './conf';
```
bartvde: ah I see now where the gitignore entry comes from

examples/mb-bright/app.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+/* eslint-disable no-underscore-dangle */
+const store = createStore(combineReducers({
+  map: SdkMapReducer,
+  print: SdkPrintReducer,
+}), window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
+   applyMiddleware(thunkMiddleware));
+
+function main() {
+  const url = 'https://raw.githubusercontent.com/mapbox/mapbox-gl-styles/master/styles/bright-v8.json';
```
bartvde: do we really want to run this from here? Or from api.mapbox.com instead?

examples/mb-bright/index.html
```diff
@@ -0,0 +1,28 @@
+<html>
+<head>
+    <title>Export Image Example</title>
```
bartvde: outdated title

examples/mb-bright/index.html
```diff
@@ -0,0 +1,28 @@
...
+    <link rel="stylesheet" type="text/css" href="/dist/sdk.css"/>
+    <link rel="stylesheet" type="text/css" href="../examples.css"/>
+</head>
+<body>
+
+    <div id="header">
+        Export Image Example
```
bartvde: outdated

examples/mb-bright/app.jsx
theduckylittle: Have a URL? This seemed like a reasonably easy source for the style sheet.

theduckylittle:
Latest update now passes all tests, good coverage, and supports everything *but* changing the sprites on the fly.

examples/mb-bright/app.jsx
bartvde: https://api.mapbox.com/styles/v1/mapbox/bright-v8

bartvde: but it will need the token as well

bartvde:
@theduckylittle is that something that needs to be added (changing sprites), or will you handle that in a new PR instead?

theduckylittle:
@bartvde  - I've added the sprite change code.

examples/mb-bright/app.jsx
theduckylittle: :+1: Awesome. :)

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules/
 coverage
 npm-debug.log
+conf.js
```
theduckylittle: Let me know if you have a better suggestion.  This is just my go-to approach.

ahocevar:
Are you guys aware of ol-mapbox-style's `apply()` function? That does the grouping already. The only thing that's currently missing is that it recognizes layers it already created when run again. That could be easily added though.

theduckylittle:
@ahocevar - I did take a look at it but had started work on this branch earlier.  I have been concerned about using `apply()`  before it had an "update cycle" working.  That said, I did update this code to use the newer `applyStyle()` function.  When the `apply()` + updates is working we should be able to quickly refactor our code.

bartvde:
Thanks for your continued work on this @theduckylittle I'll do another review now

bartvde:
Looks like something is going wrong with the sprite urls

selection_359

__tests__/components/map.test.jsx
```diff
@@ -343,15 +343,20 @@ describe('Map component', () => {
...
-    const results = jsonEquals(style, ol_style);
-    expect(results).toEqual(true);
+
+    // test that the style has been set to something
+    expect(typeof ol_style).toEqual('function');
+
+    // This is the old test.
+    // It is not a valid test as both sides will always evaluate to a function
+    // const results = jsonEquals(style, ol_style);
+    // expect(results).toEqual(true);
```
bartvde: if the old test was invalid, I would remove it altogether

__tests__/components/map.test.jsx
```diff
@@ -377,41 +382,15 @@ describe('Map component', () => {
...
-    view.getMaxResolution(), nextProps.map.layers[0].minzoom - view.getMinZoom());
-    expect(layer.getMaxResolution()).toEqual(max_rez);
-    min_rez = view.constrainResolution(
-    view.getMinResolution(), nextProps.map.layers[0].maxzoom - view.getMaxZoom());
-    expect(layer.getMinResolution()).toEqual(min_rez);
+    const layer = map.getLayers().item(0);
+
+    // a number of tests have been removed from here as the
+    // GL-layer min/maxzoom are actually handled in the style
+    // functions.
```
bartvde: no need for comments in the test IMHO, but good to know that is the reason for removal. Seems valid.

examples/mb-bright/index.html
```diff
@@ -0,0 +1,28 @@
+<html>
+<head>
+    <title>MVT Tile Example</title>
+    <link rel="stylesheet" type="text/css" href="/dist/sdk.css"/>
+    <link rel="stylesheet" type="text/css" href="../examples.css"/>
```
bartvde: please update to use the template

package.json
```diff
@@ -44,7 +45,7 @@
     "async": "^2.5.0",
     "extract-css": "^1.0.3",
     "isomorphic-fetch": "^2.2.1",
-    "mapbox-to-ol-style": "^2.5.2",
+    "mapbox-to-ol-style": "^3.1.3",
```
bartvde: is this still used separately? Or can this be removed as a dep?

src/components/map.jsx
```diff
@@ -471,56 +480,60 @@ export class Map extends React.Component {
...
    */
-  fakeStyle(layer) {
-    // getStyleFunction(glStyle, source, resolutions, spriteData, spriteImageUrl, fonts) {
-    return getStyleFunction({
+  applyStyle(olLayer, layers) {
+    // filter out the layers which are not visible
+    const render_layers = [];
+    for (let i = 0, ii = layers.length; i < ii; i++) {
+      const layer = layers[i];
+      const is_visible = layer.layout ? layer.layout.visibility !== 'none' : true;
```
bartvde: we might consider making this an util function, since e.g. the layerlist example has the same code

src/components/map.jsx
```diff
@@ -471,56 +480,60 @@ export class Map extends React.Component {
...
+    const fake_style = {
       version: 8,
-      layers: [layer],
-    }, layer.source, undefined, this.spriteData, this.spriteImageUrl);
+      sprite: this.props.map.sprite,
+      layers: render_layers,
+    };
+
+    if (olLayer.setStyle) {
+      applyStyle(olLayer, fake_style, layers[0].source, this.props.baseUrl);
```
bartvde: looks like in the case of ```mapbox://``` url sprites the baseUrl simply gets appended, see screenshot in ticket

src/components/map.jsx
```diff
@@ -284,6 +284,18 @@ function configureSource(glSource, mapProjection, accessToken) {
...
+  const all_names = [];
+  for (let i = 0, ii = layer_group.length; i < ii; i++) {
+    all_names.push(layer_group[i].id);
+  }
+  return `${layer_group[0].source}-${all_names.join(',')}`;
+}
+
+/** Convert a zoom number to a resolution.
+ */
 function getResolutionForZoom(map, zoom) {
```
bartvde: since we are upgrading ol to 4.3.1 anyway we can use http://openlayers.org/en/latest/apidoc/ol.View.html#getResolutionForZoom ?

webpack-dev-server.config.js
```diff
@@ -65,6 +65,10 @@ const config = {
       'webpack/hot/only-dev-server',
       './examples/rotating/app.jsx',
     ],
+    'mb-bright': [
+      'webpack/hot/only-dev-server',
```
bartvde: not needed anymore after you pull from master for the example templates

package.json
ahocevar: This is not needed any more.

ahocevar: Can definitely be removed.

src/components/map.jsx
ahocevar: You need to pass the full url with access token here, not the base url.

bartvde: Just tried this with: ```applyStyle(olLayer, fake_style, layers[0].source, `${this.props.baseUrl}?access_token=${this.props.accessToken}`);``` however still seeing some issues

bartvde: selection_360

bartvde: Ah maybe because the baseUrl seems to be for v9 and the style doc is loaded from v8

bartvde: No that doesn't seem to make a difference

ahocevar: I just saw that this is a bug in ol-mapbox-style. Access tokens are handled in `apply()` only, not in `applyStyle()`.

ahocevar: Well, it's not exactly a bug. applyStyle does not know about the access token. So we'd either need to change its API, or you have to modify the sprite url in the glStyle object before calling applyStyle:
```js
  if (glStyle.sprite && glStyle.sprite.indexOf('mapbox://') == 0) {
    glStyle.sprite = baseUrl + '/sprite?access_token=' + accessToken;
  }
```
```

bartvde: So the path in applyStyle is only used to make relative urls absolute right? I think it would be ok to handle this on our side.

ahocevar: Right. The snipped I provided above should handle things on the app side.

bartvde: Hmm thinking about it again, it might not be so convenient because we would not serialize back exactly what we got. Let's see what @theduckylittle says when he gets online today.

ahocevar: Well, apply() does the same - it changes the urls in the glStyle object. So maybe a better approach would be to change ol-mapbox-style's API so users need to provide an access token when using mapbox:// resources.

bartvde: Yes that makes sense to me.

bartvde:
@theduckylittle I wouldn't worry about the decrease in coverage too much, I'd be happy to work on increasing it again after this gets merged

theduckylittle:
Alright, y'all updated this PR.

I know this has a lot of commits against it but I'll squash them during the merge once everything checks out.

src/components/map.jsx
theduckylittle: This logic is used in a number of places.  I think we should file a follow up ticket that deals with that and tries to put it in the appropriate places in the code.

theduckylittle:
This push fixes the "flashing" and constant redrawing.

theduckylittle:
Ack, do not merge, I found a separate bug with the ref layers, will push again once it's fixed.

PR #614, opened 2017-08-18
Update instructions for node-canvas package dependencies

travislbrundage:


bartvde:
Perhaps we should link to https://github.com/boundlessgeo/sdk/blob/master/DEVELOPING.md instead?

travislbrundage:
I didn't notice that file actually, but I get the error not when testing, but trying to install in the first place. I cannot successfully `npm install` before installing those dependencies, so to me it would make sense to put it where the install instructions are.

theduckylittle:
`canvas` should be removed from the devDependencies...

theduckylittle:
Merged #626 , this should keep the installer from trying to put canvas locally.  Tests won't run appropriately but you can USE SDK without it.

PR #615, opened 2017-08-18
[WIP] Upgrade to React 16

bartvde:
I'm only opening up this now so we have a good view of the changes.
We need to wait until React 16 is really released and when Enzyme has a release that is compatible with it.
The changes needed in our code base seem small.


bartvde:
Will close this for now, will revisit later

PR #616, opened 2017-08-18
Minor fixes to the examples

bartvde:
This came up during #615 which showed warnings / errors for these issues


bartvde:
Example only so merging.

PR #617, opened 2017-08-18
Make sure we reload layers/sources if version is undefined

bartvde:
SDK-638


bartvde:
This is now ready for review

PR #618, opened 2017-08-22
Adding optional parameter to specify cross origin requests

travislbrundage:


src/services/MapConfigTransformService.js
```diff
@@ -183,7 +183,11 @@ class MapConfigTransformService {
       thumbnail: thumbnail
     };
   }
-  transform(data, opt_errors, opt_tileServices) {
+  transform(data, opt_errors, opt_tileServices, opt_crossOrigin) {
```
bartvde: I'd either pass in a string (so ```use-credentials``` as the value), or rename this to opt_useCredentials

src/services/MapConfigTransformService.js
```diff
@@ -235,15 +239,15 @@ class MapConfigTransformService {
         layerConfig.source = {
           type: 'OSM',
           properties: {
-            crossOrigin: 'anonymous'
+            crossOrigin: crossOrigin
```
bartvde: I would say for OSM sources this will never be needed? Since their tiles will remain freely accessible?

travislbrundage:
I updated the changes as you requested, thank you for reviewing!

travislbrundage:
One moment, small mistake is happening - fixing

PR #619, opened 2017-08-22
Templating our examples

bartvde:
SDK-631

The approach used was borrowed from the openlayers project


bartvde:
This is ready for review now

bartvde:
I realised I need to see how I can run tasks/build-examples.js on demand when needed

bartvde:
Although it currently also does not watch for html changes

tasks/build-examples.js
```diff
@@ -0,0 +1,123 @@
+var path = require('path');
```
wnordmann: This file ins't being linted, the`var` is a give away

tasks/build-examples.js
```diff
@@ -0,0 +1,123 @@
...
+function createIndex(files, metalsmith, done) {
+  setImmediate(done); // all remaining code is synchronous
+  var index = '<!DOCTYPE html>';
+  index += '<html>';
+  index += '<head>';
+  index += '<title>Boundless SDK Examples</title>';
+  index += '<link rel="stylesheet" type="text/css" href="examples.css"/>';
+  index += '</head>';
+  index += '<body>';
+  index += '<div id="header">';
```
wnordmann: A link back to `examples/index.html` would be good

bartvde:
thanks for the review @wnordmann working on the changes now

tasks/build-examples.js
bartvde: I agree this would be good, but I think it would make sense for @jjmulenex to review our example template, maybe we should add a navbar or something. We can incorporate it in that review easily IMHO.

bartvde: Btw, I think you meant this comment to be for the template, not the index page right?

PR #620, opened 2017-08-22
Change eslint config to use config-react-app

bartvde:
This change is needed so we can use SDK within create-react-app


theduckylittle:
Something like...

```bash
find ./__tests__ -name "*.jsx" -exec /bin/bash -c 'git mv "$1" "${1%.jsx}".js' _ {} \;
find ./examples -name "*.jsx" -exec /bin/bash -c 'git mv "$1" "${1%.jsx}".js' _ {} \;
```

bartvde:
@theduckylittle PR updated

PR #622, opened 2017-08-23
 First Attempt at Packaging and Documentation

theduckylittle:
Didn't know whether to label this as WIP or as being for critical review. Here's what's going on:

1. Uses babel-cli to create a `./lib` directory.
2. All of the examples have been updated to point to the `./lib` directory
3. webpack-dev-server will actually forward `lib` to `src`
   so development should remain seamless.
4. Added documentation for getting started via create-react-scripts. (More readable form here: https://github.com/theduckylittle/sdk/blob/packaging/phase-1/docs/getting-started.md)


examples/basic-wgs84/app.js
```diff
@@ -11,10 +11,10 @@ import thunkMiddleware from 'redux-thunk';
 import React from 'react';
 import ReactDOM from 'react-dom';
 
-import SdkMap from '@boundlessgeo/sdk/components/map';
-import SdkHashHistory from '@boundlessgeo/sdk/components/history';
-import SdkMapReducer from '@boundlessgeo/sdk/reducers/map';
-import * as mapActions from '@boundlessgeo/sdk/actions/map';
+import SdkMap from '@boundlessgeo/sdk/lib/components/map';
```
bartvde: personally I like imports without any "root" path in them, we can transpile to lib and publish from there so that the lib addition is not needed

bartvde: also, this was feedback given to us during SDK 1.x by other devs.  We used to have a path in between as well and we took it out.

examples/export-image/app.js
```diff
@@ -25,6 +25,8 @@ const store = createStore(combineReducers({
 }), window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
    applyMiddleware(thunkMiddleware));
 
+window.store = store;
```
bartvde: not intentional I guess ;-) ?

examples/export-image/app.js
```diff
@@ -37,6 +39,7 @@ function main() {
   // place the map on the page.
   ReactDOM.render(
     <SdkMap
+      ref={(m) => { window.map = m; }}
```
bartvde: not intentional I guess ;-)

package.json
```diff
@@ -8,9 +8,11 @@
     "test:watch": "npm test -- --watch",
     "lint": "eslint __tests__ src examples tasks",
     "cover": "node_modules/.bin/jest --coverage",
-    "build:examples": "node tasks/build-examples.js"
+    "build:examples": "node tasks/build-examples.js",
+    "build:js": "babel ./src --out-dir ./lib"
```
bartvde: I guess we should be setting things like NODE_ENV here as well?

PR #624, opened 2017-08-24
Bring test coverage back up

bartvde:
SDK-645


bartvde:
tests only so merging this one

PR #626, opened 2017-08-24
Drop Canvas from devDeps

theduckylittle:
Installing the canvas module has system requirements that
are not universally necessary for INSTALLING SDK.  They are
really only useful for running the test suite.

The tests which use canvas have checks to see if it is installed
making it, essentially, an optional install.


bartvde:
what does this mean for our tests on Travis and coverage?

theduckylittle:
i have some incremental commits to clean up during the merge but now Travis + Coveralls seems to be working appropriately.

PR #629, opened 2017-08-25
App/load mapbox style

CarlyLanglois:
app to render map via url to mapbox style doc or mapbox style doc file upload


examples/mb-bright/app.js
```diff
@@ -16,7 +16,7 @@ import * as mapActions from '@boundlessgeo/sdk/actions/map';
 // This will have webpack include all of the SDK styles.
 import '@boundlessgeo/sdk/stylesheet/sdk.scss';
 
-import CONFIG from './conf'; // eslint-disable-line
+// import CONFIG from './conf'; // eslint-disable-line
```
CarlyLanglois: FYI i had to comment this out to avoid a compiler error w/ this ex app

bartvde: what compiler error exactly? The idea is that you place a file in examples/mb-bright with your mapbox api key, something like:
```
const config = {
  access_token: 'your_key_here'
};

export default config;
```

CarlyLanglois: ok that works. i wasnt able to run npm start since I didnt have the conf file in the app dir

examples/upload-map/context-selector.js
```diff
@@ -0,0 +1,136 @@
...
+      this.props.setContext({ json: context.json });
+    } else {
+      this.props.setContext({ url: context.url });
+    }
+  }
+
+  dropFiles(files) {
+    this.setState({ error: false });
+    this.message = '';
+    this.badurl = '';
```
bartvde: we can probably combine these 3 lines in a helper function since they are used multiple times

examples/upload-map/context-selector.js
```diff
@@ -0,0 +1,136 @@
...
+          result = JSON.parse(r.result);
+          this.addMap({ json: result });
+        } catch(e) {
+          this.message = e.message;
+          this.setState({ error: true });
+        }
+      };
+      r.onerror = () => {
+        this.message = r.error;
+        this.setState({ error: true });
```
bartvde: same for these 2 lines, can be a helper function ```setError(message)```

examples/upload-map/context-selector.js
```diff
@@ -0,0 +1,136 @@
...
+      };
+      r.readAsText(file);
+    }
+  }
+
+  handleChange(event) {
+    this.setState({ error: false });
+    this.message = '';
+    this.badurl = '';
+    this.setState({ value: event.target.value });
```
bartvde: I'd use a single setState call here, not 2

examples/upload-map/context-selector.js
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  handleSubmit(event) {
+    this.setState({ error: false });
+    this.message = '';
+    try {
+      fetch (this.state.value)
```
bartvde: probably good to import fetch in this file
```
import fetch from 'isomorphic-fetch';
```

examples/upload-map/context-selector.js
```diff
@@ -0,0 +1,136 @@
...
+        .then(
+          response => {
+            if (response.status >= 400) {
+              console.log(response)
+              this.badurl = response.url
+              this.message = response.status + ' (' + response.statusText + ')';
```
bartvde: in other places were use ES6 templates, but maybe our new eslint rules do not ask this anymore

examples/upload-map/context-selector.js
```diff
@@ -0,0 +1,136 @@
...
+              this.setState({ error: true });
+            } else {
+              this.addMap({ url: this.state.value });
```
bartvde: since we have the content already here, why are we not passing on the json instead of having SDK having the reload the url again?

examples/upload-map/context-selector.js
```diff
@@ -0,0 +1,136 @@
...
+        this.message = e
+        this.setState({ error: true });
+    }
+    event.preventDefault();
+  }
+
+  render() {
+    let errormsg;
+    if (this.state.error === true) {
+      errormsg = (<p className="error"><strong>Error uploading map:</strong> <a href={this.badurl}>{this.badurl}</a> {this.message}</p>)
```
bartvde: I'd probably put message and badurl in the component's state as well, since it's used for rendering?

examples/upload-map/context-selector.js
```diff
@@ -0,0 +1,136 @@
...
+        <div className="selector">
+          <div className="url">
+            <h2>Via URL: </h2>
+            <form onSubmit={this.handleSubmit}>
+              <input className="dropzone" id="urlField" placeholder="Provide a URL and hit enter to render map above" type="text" value={this.state.value} onChange={this.handleChange} />
+            </form>
+          </div>
+          <div className="drop">
+            <h2>Via File Upload:</h2>
+            <Dropzone className="dropzone" accept=".json,.geojson" multiple={false} onDrop={this.dropFiles}>
```
bartvde: geojson doesn't make a lot of sense as extension here probably?

package.json
```diff
@@ -59,6 +59,7 @@
     "prop-types": "^15.5.10",
     "react": "^15.6.1",
     "react-dom": "^15.6.1",
+    "react-dropzone": "^4.0.1",
```
bartvde: since this is not a dependency of sdk, this shouldn't be here
We should use unpkg.com instead with a script tag per my earlier comment

examples/upload-map/context-selector.js
CarlyLanglois: I'm not sure i quite understand the question here. is it why are we adding the map using the url and not the json contents? If so, I don't actually have a good answer for you, so maybe worth changing..

bartvde:
right that is the question, I would recommend passing the json instead of the url here

examples/upload-map/context-selector.js
```diff
@@ -0,0 +1,132 @@
...
+    try {
+      fetch (this.state.value)
+        .then(
+          response => {
+            if (response.status >= 400) {
+              this.setState({ badurl: response.url });
+              const msg = `${response.status} (${response.statusText})`;
+              this.setError(msg);
+            } else {
+              this.addMap({ url: this.state.value });
```
CarlyLanglois: thanks bart! I had a question on your comment on this line but it got buried when I pushed changes. 
Were you asking why not set context w/ json here instead of the url since we're already fetching its contents?

bartvde:
indeed, exaxtly

CarlyLanglois:
ok, yeah no reason other than setContext can handle a url argument itself. I only brought the fetch in when working on error handling but I've made a change to use the json contents now

bartvde:
@carlylanglois only thing that needs to be done is remove the import for react-dropzone, after that feel free to merge

CarlyLanglois:
removed Dropzone for input type=file

CarlyLanglois:
JIRA SDK-588

PR #633, opened 2017-08-29
WFS-T

theduckylittle:
Initial example using WFS-T to edit a layer with SDK.

This version has the new WFS controller embedded into
the example and not moved into the library.

*Warning:* This includes PR #632 for which I will have tests written for soon.


examples/wfst/wfs-controller.js
```diff
@@ -0,0 +1,124 @@
...
+      const working_srs = 'EPSG:3857';
+
+      // add it to the queue
+      this.pendingActions[id] = action;
+
+      const src = props.sources[action.sourceId];
+
+      const wfs_format = new WfsFormat();
+      const geojson_format = new GeoJsonFormat();
+
```
bartvde: I'd suggest creating the formats once outside of this function

examples/wfst/wfs-controller.js
```diff
@@ -0,0 +1,124 @@
...
+      const json_feature = jsonClone(action.feature);
+      delete json_feature.properties['bbox'];
+
+      const feature = geojson_format.readFeature(json_feature, {
+        dataProjection: 'EPSG:4326',
+        featureProjection: working_srs,
+      });
+
+      let geom_name = 'geometry';
+      if (src.metadata['bnd:geometryName'] !== undefined) {
```
bartvde: shouldn't metadata be on the layer according to the mapbox style spec since there is no metadata on the source level in the spec? I thought we had moved it to the map's metadata instead with a hash with the sourceName?

examples/wfst/wfs-controller.js
```diff
@@ -0,0 +1,124 @@
...
+
+      // convert this to a WFS call.
+      const xml = wfs_format.writeTransaction(
+        actions.insert, actions.update, actions.delete, options);
+
+      // convert the XML to a string.
+      let payload = (new XMLSerializer()).serializeToString(xml);
+
+      // fix the geometry name...
+      payload = payload.replace('<Name>geometry</Name>', `<Name>${geom_name}</Name>`);
```
bartvde: can't we change the feature's geometryName instead of changing the output from the format?

examples/wfst/wfs-controller.js
```diff
@@ -0,0 +1,124 @@
...
+
+      // get the target_url from the service
+      const target_url = src.data.split('?', 1)[0];
```
bartvde: I'm guessing this might not work for all use cases. Might be better to put the OnlineResource in the metadata instead?

examples/wfst/wfs-controller.js
```diff
@@ -0,0 +1,124 @@
...
+      //   to the queue.
+      //  if there is an error (bad request, etc.) then post an error
+      //  if the request completes cleanly dispatch a refresh on that source.
+      fetch(target_url, {
+        method: 'POST',
+        body: payload,
+      }).then((response) => {
+        return response.text();
+      }).then((text) => {
+        const response_xml = (new DOMParser()).parseFromString(text, 'text/xml');
```
bartvde: can't we let the WFS format from openlayers parse the response? See: http://openlayers.org/en/latest/apidoc/ol.format.WFS.html#readTransactionResponse

examples/wfst/wfst.js
```diff
@@ -0,0 +1,53 @@
+
+import uuid from 'uuid';
```
bartvde: maybe put this file in a reducers subdir so it's clear that it's a reducer

examples/wfst/wfst.js
```diff
@@ -0,0 +1,53 @@
...
+    // remove a source from the WFS configuration
+    case 'remove-source':
+      return removeSource(state, action);
+    case 'wfs:insert':
+    case 'wfs:update':
+    case 'wfs:delete':
+      console.log('add action', state, action, addAction(state, action));
+      return addAction(state, action);
+    case 'finished':
+      return finishAction(state, action);
```
bartvde: there is no finishAction?

examples/wfst/app.js
```diff
@@ -0,0 +1,154 @@
...
+    ],
+  }));
+
+  const tracts_url = '/geoserver/wfs?SRSNAME=EPSG:3857&REQUEST=GetFeature&TYPENAME=sdk:minnesota_tracts&VERSION=1.1.0&OUTPUTFORMAT=JSON';
+
+  store.dispatch(SdkMapActions.addSource('tracts', {
+    type: 'geojson',
+    data: tracts_url,
+    metadata: {
+      'bnd:ns': 'sdk',
```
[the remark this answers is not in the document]
theduckylittle: Normally, yes, but when I looked at the xml version it had `xmlns:sdk="sdk"`.  So I went with it.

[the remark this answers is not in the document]
bartvde: It won't be valid XML probably

examples/wfst/app.js
```diff
@@ -0,0 +1,154 @@
...
+    store.dispatch(SdkMapActions.addLayer({
+      id: `tracts-${i}`,
+      type: 'fill',
+      source: 'tracts',
+      paint: {
+        'fill-color': colors[i],
+        'fill-opacity': 0.5,
+        'fill-outline-color': 'black',
+      },
+      filter: ['==', 'color', i],
```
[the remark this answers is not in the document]
theduckylittle: Yes, but I'm not 100% confident how we'll implement that yet. I find that to be a fundamental limitation to the MapBox Styles format.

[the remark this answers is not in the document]
bartvde: I'm guessing we could handle it the same way as the WMS GetMap with the bbox substitute?

examples/wfst/app.js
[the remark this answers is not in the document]
theduckylittle: I just do what GeoServer tells me to do. ;)

examples/wfst/app.js
```diff
@@ -0,0 +1,154 @@
...
+    type: 'raster',
+    tileSize: 256,
+    tiles: [
+      'https://a.tile.openstreetmap.org/{z}/{x}/{y}.png',
+      'https://b.tile.openstreetmap.org/{z}/{x}/{y}.png',
+      'https://c.tile.openstreetmap.org/{z}/{x}/{y}.png',
+    ],
+  }));
+
+  const tracts_url = '/geoserver/wfs?SRSNAME=EPSG:3857&REQUEST=GetFeature&TYPENAME=sdk:minnesota_tracts&VERSION=1.1.0&OUTPUTFORMAT=JSON';
```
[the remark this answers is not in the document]
theduckylittle: We could, in theory, implement a similar strategy as is used by the WMS BBOX.

theduckylittle:
WRT: "What's in the demo vs What's in Core", yes, the reducer and the 'controller' will be put into the core.  I started it this was just for my own convenience and also wanted a clean way to tell what needed to change in the existing core to support these changes.  As we're getting increasingly stable and increasingly good coverage I didn't want to have something go sideways with this PR.

bartvde:
cool makes sense @theduckylittle with playback I'm also experimenting the same way to see what needs to be in core and what doesn't need to be there

examples/wfst/app.js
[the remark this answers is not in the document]
bartvde: GeoServer will not disallow it probably, but run it through any XML validator and it will complete if xmlns is not a URI

examples/wfst/app.js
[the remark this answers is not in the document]
bartvde: My comment should have gone above, wrt the WMS BBOX strategy, you suggested the same actually.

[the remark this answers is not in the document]
bartvde: We could probably handle it the same way as Andreas does for the vector tile stuff.

examples/wfst/app.js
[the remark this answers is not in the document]
theduckylittle: Yes, but the XML *returned from GeoServer* contains `xmlns:sdk="sdk"`, so this configuration bit is really just matching what the server is looking for, that said, this isn't in the controller -- so it's not a huge problem to change on an app-by-app basis.

examples/wfst/app.js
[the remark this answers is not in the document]
theduckylittle: My only objection is that Mapbox does not actually specify a way of handling a "geojson" + BBOX layer.

examples/wfst/wfs-controller.js
theduckylittle: It does't work in the current openlayers code. The `getKeys` function returns `geometry` no matter what it's named in the feature.

examples/wfst/wfs-controller.js
theduckylittle: converted to defining the online resource explicitly.

examples/wfst/wfs-controller.js
theduckylittle: I moved all the "metadata" into the wfs reducer.  That way it's removed from the mapbox definition.

examples/wfst/wfs-controller.js
theduckylittle: Put them in the constructor.

examples/wfst/wfs-controller.js
theduckylittle: Converted to using that function instead of the manual parsing.

theduckylittle:
@bartvde  I think I have all of your requested changes. I know this is missing tests but let me know if you see anything I missed!

examples/wfst/app.js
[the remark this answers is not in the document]
bartvde: Ah then you've configured your GeoServer wrongly :-)

examples/wfst/app.js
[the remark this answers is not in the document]
bartvde: I know, but I think we are free to come up with extension points if needed. We should check it doesn't break when we load our config into another piece of software.

examples/wfst/wfs-controller.js
bartvde: the code should use getGeometryName() and I've seen that work fine in the past. Can you check the outcome of getGeometryName? Maybe the geometry name is not being set correctly on the feature?

bartvde:
@theduckylittle I'd like to get to the bottom of the issue on the geometry name if possible, instead of the workaround. I'll have another look at the code today.

examples/wfst/app.js
```diff
@@ -0,0 +1,155 @@
+/** Very basic SDK application example.
+ *
+ *  Contains a Map and demonstrates some of the dynamics of
+ *  using the store.
+ *
```
bartvde: We should update this section

examples/wfst/app.js
```diff
@@ -0,0 +1,155 @@
...
+import SdkDrawingReducer from '@boundlessgeo/sdk/reducers/drawing';
+import * as SdkDrawingActions from '@boundlessgeo/sdk/actions/drawing';
+import * as SdkWfsActions from '@boundlessgeo/sdk/actions/wfs';
+
+import { WfsReducer } from '@boundlessgeo/sdk/reducers/wfs';
+
+import WfsController from '@boundlessgeo/sdk/components/wfs';
+
+// MapBox GL Styles does not support WFS layers.
+// import SdkWfsReducer from '@boundlessgeo/sdk/reducers/wfs';
```
bartvde: please remove commented line

examples/wfst/app.js
```diff
@@ -0,0 +1,155 @@
...
+
+/* eslint-disable no-underscore-dangle */
+const store = createStore(combineReducers({
+  map: SdkMapReducer,
+  drawing: SdkDrawingReducer,
+  wfs: WfsReducer,
+}), window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
+   applyMiddleware(thunkMiddleware));
+
+window.store = store;
```
bartvde: should take this out before the final PR

examples/wfst/app.js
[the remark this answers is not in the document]
bartvde: We can leave BBOX handling out for now. Maybe create a JIRA though so we don't forget to revisit.

examples/wfst/app.js
```diff
@@ -0,0 +1,155 @@
...
+  const tracts_url = '/geoserver/wfs?SRSNAME=EPSG:3857&REQUEST=GetFeature&TYPENAME=sdk:minnesota_tracts&VERSION=1.1.0&OUTPUTFORMAT=JSON';
+
+  store.dispatch(SdkMapActions.addSource('tracts', {
+    type: 'geojson',
+    data: tracts_url,
+  }));
+
+  // this will configure the source for WFS editing.
+  store.dispatch(SdkWfsActions.addSource('tracts',
+    '/geoserver/wfs',
```
bartvde: does this require any proxy handling on the webpack dev server?

examples/wfst/app.js
```diff
@@ -0,0 +1,155 @@
...
+    'sdk',
+    'sdk',
+    'minnesota_tracts',
+    'wkb_geometry',
```
bartvde: I wonder if we should change this to a config object, instead of a list of strings, to make it more descriptive?

examples/wfst/app.js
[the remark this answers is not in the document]
bartvde: Fine to leave this out for now, let's at least create a JIRA so we can revisit if needed

examples/wfst/index.html
```diff
@@ -0,0 +1,10 @@
+---
+layout: example.html
+title: WFS-T Map Editing Example
+shortdesc: Show how to work with basic WFS-T editing.
+docs: >
+    <li>Add a layer as WMS</li>
+    <li>Use WMS GetFeatureInfo results to send a WFS-T request</li>
```
bartvde: is this accurate? Where is WMS GetFeatureInfo involved?

examples/wfst/index.html
```diff
@@ -0,0 +1,10 @@
+---
+layout: example.html
+title: WFS-T Map Editing Example
+shortdesc: Show how to work with basic WFS-T editing.
+docs: >
+    <li>Add a layer as WMS</li>
```
bartvde: There is no WMS layer right?

src/actions/map.js
```diff
@@ -211,3 +211,15 @@ export function updateMetadata(metadata) {
...
   };
 }
+
+
+/** Manually update a source.
+ */
+export function updateSource(sourceId, update) {
+  return {
+    type: MAP.UPDATE_SOURCE,
+    sourceId,
```
bartvde: not very important, but I'd use sourceName to be consistent as the var name

src/components/wfs.js
```diff
@@ -0,0 +1,125 @@
...
+    this.geojson_format = new GeoJsonFormat();
+  }
+
+  execute(props, id) {
+    // only act if the action is not already pending.
+    if (this.pendingActions[id] === undefined) {
+      // copy the action
+      const action = Object.assign({}, props.actions[id]);
+
+      const working_srs = 'EPSG:3857';
```
bartvde: I guess this should not be hard-coded eventually?

src/components/wfs.js
```diff
@@ -0,0 +1,125 @@
...
+
+      // clone the feature, as GeoJSON features have a lot of
+      //  depth this ensures all the sub-objects are cloned reasonably.
+      const json_feature = jsonClone(action.feature);
+      delete json_feature.properties['bbox'];
+
+      const feature = this.geojson_format.readFeature(json_feature, {
+        dataProjection: 'EPSG:4326',
+        featureProjection: working_srs,
+      });
```
bartvde: this is probably where we do not get the right geometry name for our feature created. The GeoJSON format can get an optional geometryName, maybe we should utilize this to get the correct feature with the correct geom name. It would mean moving the format out of the constructor again, but so be it ;-)

src/components/wfs.js
```diff
@@ -0,0 +1,125 @@
...
+        actions[WFS.INSERT],
+        actions[WFS.UPDATE],
+        actions[WFS.DELETE],
+        options);
+
+      // convert the XML to a string.
+      let payload = (new XMLSerializer()).serializeToString(xml);
+
+      // fix the geometry name...
+      payload = payload.replace('<Name>geometry</Name>', `<Name>${geom_name}</Name>`);
```
bartvde: see my comment above for specifying geometryName in the GeoJSON format constructor instead

src/components/wfs.js
```diff
@@ -0,0 +1,125 @@
...
+
+      // get the target_url from the service
+      const target_url = src.onlineResource;
+
+      // attempt the action,
+      //  if there is a failure (offline) then add the action back
+      //   to the queue.
+      //  if there is an error (bad request, etc.) then post an error
+      //  if the request completes cleanly dispatch a refresh on that source.
+      fetch(target_url, {
```
bartvde: we should probably import fetch in this file just in case?

bartvde: also one of the things we haven't done in SDK 2.x, but which will be important for the new geonode client is specifying a proxy for XHR

src/components/wfs.js
```diff
@@ -0,0 +1,125 @@
...
+
+  render() {
+    // never render anything.
+    return false;
+  }
+}
+
+function mapStateToProps(state) {
+  return {
+    actions: state.wfs.actions, //(state.wfs && state.wfs.actions) || null,
```
bartvde: can we get rid of the comment?

src/reducers/map.js
```diff
@@ -389,6 +389,24 @@ function updateMetadata(state, action) {
...
 }
 
+/** Update a source's definition.
+ *
+ *  This is a heavy-handed operation that will
+ *  just mixin whatever is in the new object.
+ *
+ */
+function updateSource(state, action) {
+  const old_source = state.sources[action.sourceId];
```
bartvde: I'd use sourceName as stated above

src/reducers/map.js
```diff
@@ -389,6 +389,24 @@ function updateMetadata(state, action) {
...
+  const new_source = {};
+  new_source[action.sourceId] = Object.assign({}, old_source, action.sourceDef);
+  const new_sources = Object.assign({}, state.sources, new_source);
+  //return Object.assign({}, state, {sources: new_sources});
```
bartvde: can we get rid of the comment?

webpack-dev-server.config.js
```diff
@@ -36,6 +36,12 @@ const config = {
     inline: true,
     port: 3000, // Port Number
     host: 'localhost', // Change to '0.0.0.0' for external facing server
+    proxy: {
+      '/geoserver' : {
+        target: 'http://localhost:8080/',
```
bartvde: ah there is my answer to the proxy question

examples/wfst/app.js
theduckylittle: Some day, I might conceed to actually installing the Redux devtools extension.  Today is not that day. LONG LIVE THE CONSOLE!

examples/wfst/index.html
theduckylittle: correct.

src/actions/map.js
theduckylittle: Yeah, I'll fix that in my other PR.

examples/wfst/index.html
theduckylittle: This was my original plan and then I got the WFS-T stuff working and ignored that plan.

src/components/wfs.js
theduckylittle: This was a cheat. When using WFS and EPSG:4326, it requires that the coordinates get flipped but OL wasn't doing that automatically so I switched to a meters-based projection that is consistently X,Y.

bartvde: Can you try something like: https://github.com/boundlessgeo/suite/blob/r4.8/sdk/src/main/resources/client/ol3view/src/app/app.js#L23:L28

src/components/wfs.js
theduckylittle: The problem is in `_ol_format_WFS_.writeUpdate_`, it uses `feature.getKeys()`.  `feature.getKeys()` ALWAYS returns `geometry` as the name of the geometry column. But `_ol_format_GML3_.prototype.writeFeatureElement` (used by Insert/writeFeature) will actually use `feature.getGeometreyName`.

src/reducers/map.js
theduckylittle: will fix in the other PR (expect that to flip from WIP first!)

src/components/wfs.js
theduckylittle: Implemented.

src/components/wfs.js
bartvde: ok then make sure we ticket this in ol, maybe with a failing test

src/components/wfs.js
theduckylittle: Added fetch.

src/components/wfs.js
theduckylittle: I'll work on that tomorrow. I'll be burning a lot of time today writing tests for the rest of the stack.

__tests__/components/wfs.test.js
```diff
@@ -0,0 +1,122 @@
...
+      onRequestError: () => {
+      },
+    };
+    spyOn(props, 'onRequestError');
+
+    mount(<WfsController {...props} />);
+
+    store.dispatch(actions.addSource('test', {
+      onlineResource: 'http://example.com/wfs',
+      xmlNS: 'http://example.com/sdk',
```
bartvde: I'd prefer ```featureNS``` here but not crucial

__tests__/components/wfs.test.js
```diff
@@ -0,0 +1,122 @@
...
+
+    nock('http://fail.com')
+      .post()
+      .reply(500);
+    store.dispatch(actions.updateFeature('fail-source', feature));
+
+    setTimeout(() => {
+      // if everything has gone well the action will no longer be in the queue
+      expect(store.getState().wfs.actions).toEqual({});
+      // ensuire the error handler was called.
```
bartvde: ensure

examples/wfst/app.js
```diff
@@ -0,0 +1,153 @@
...
+
+/* eslint-disable no-underscore-dangle */
+const store = createStore(combineReducers({
+  map: SdkMapReducer,
+  drawing: SdkDrawingReducer,
+  wfs: SdkWfsReducer,
+}), window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
+   applyMiddleware(thunkMiddleware));
+
+window.store = store;
```
bartvde: it's back :-)

theduckylittle:
Refs: SDK-621.

Followup work filed in SDK-661

PR #634, opened 2017-08-29
Adding in new images to support high res screens

wnordmann:
https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio


CarlyLanglois:
JIRA SDK 659

examples/sprites/sprites@2x.json
```diff
@@ -0,0 +1,23 @@
+{
+  "duck": {
+    "height": 32,
+    "pixelRatio": 1,
```
bartvde: shouldn't pixelRatio be 2 in the 2.x json file?

PR #635, opened 2017-08-30
Adding export map action to Upload App

CarlyLanglois:
save button saves map spec to .json file - i added it to this app so you can demo saving a spec file, altering it, and re-uploading it into the app.

JIRA SDK 554 - the jira ticket mentions wmc format but this button saves the file as .json, per talks w/ Walker and the team. 

Right now the whole map state is saved but if less or more info is needed for the map state - ie, state.map.sprite? - let me know.

### DIFF
--- a/examples/upload-map/app.js
+++ b/examples/upload-map/app.js
@@ -53,7 +53,7 @@ function main() {
     <div>
       <ContextSelector store={store} />
       <h1>Save a Map</h1>
-      <h2>To .json:</h2>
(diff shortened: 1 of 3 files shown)


examples/upload-map/app.js
```diff
@@ -44,6 +52,9 @@ function main() {
   ReactDOM.render((
     <div>
       <ContextSelector store={store} />
+      <h1>Save a Map</h1>
+      <h2>To .json:</h2>
```
bartvde: I would probably add to MapBox Style Specification and link to https://www.mapbox.com/mapbox-gl-js/style-spec/

PR #638, opened 2017-09-05
deleting getMin and Max util functions that are no longer used

CarlyLanglois:
getMin and getMax functions have been previously replaced w/ ol getResolutionForZoom function


bartvde:
those functions were probably 100% tested @CarlyLanglois that's why the average went down

PR #648, opened 2017-09-07
removing a removePopup() method not used

CarlyLanglois:
popups are removed via the updatePopups() method

### DIFF
--- a/__tests__/components/map.test.js
+++ b/__tests__/components/map.test.js
@@ -823,25 +823,6 @@ describe('Map component', () => {
     expect(sdk_map.updatePopups).toHaveBeenCalled();
   });
 
-  it('should remove a popup', () => {
-    const store = createStore(combineReducers({
-      map: MapReducer,
-    }));
-
-    const props = {
-      store,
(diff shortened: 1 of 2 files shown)


bartvde:
hmm does this make sense? I think we would still want a way for people to programatically remove a popup? cc @theduckylittle

CarlyLanglois:
oh good point. but if the map component accomplishes removing a popup w/o this method, would it make sense to just include a remove function inside of whatever app needs it?

bartvde:
it is true that popups can still be closed by calling the close method on the popup itself

I'm not sure if we need to have 2 ways, I'll let @theduckylittle decide since he wrote the original popup code

theduckylittle:
Please don't remove potentially useful API calls. :)

`updatePopups` does the real work but we should have an exposed method for removing a popup by ID.  The only real flaw here is that we should be returning the popup ID in addPopup so it's easier to reference.

CarlyLanglois:
ok, thanks for the notes. I'll close this then

PR #649, opened 2017-09-07
prepping notes and comments for jsdoc generation

CarlyLanglois:
JS Doc coverage on src files - components, actions, reducers, and helper functions.


CarlyLanglois:
npm run jsdoc generates a folder of doc files to browse, including the getting started readme info on its landing page. I think the styling could use some improvements to more easily view each of the methods.

Also, is it more appropriate to include jsdoc as a devDependency than a dependency? And is it the right more to gitIgnore the output folder? Thanks!

bartvde:
in openlayers jsdoc is a dependency: https://github.com/openlayers/openlayers/blob/master/package.json#L43

bartvde:
ignoring the output dir makes sense to me

CarlyLanglois:
thanks bart. I just took it out of this pr but I see there's also a lot of extraneous templating files that I think are required in order for the custom template to be used. is there a better way to handle these rather than committing them too?

bartvde:
I'm not sure, I would advise you to check how it's done in openlayers

theduckylittle:
The re-org looks great. Github is reporting a merge conflict, once this has been updated to master, I think it's good to merge --- pending any template changes you'd like to make.  This is at least *something* and a good starting point.

PR #650, opened 2017-09-08
Feature/examples page

jjmulenex:
updated layout to make it more consistent across sample pages

### DIFF
--- a/config/examples/example.html
+++ b/config/examples/example.html
@@ -11,14 +11,16 @@
     <div id="header">
       {{ shortdesc }}
     </div>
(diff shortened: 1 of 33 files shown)


config/examples/example.html
```diff
@@ -9,14 +9,20 @@
   </head>
   <body>
     <div id="header">
-      {{ shortdesc }}
+      <a href="/build/examples"><img src="../boundless_sdk_horiz.svg" width="90"></a>
```
bartvde: can we use a relative url instead?

PR #651, opened 2017-09-08
Filterbuilder532

wnordmann:
Simple app to testing filter functions based on mapbox styles


bartvde:
instead of removing from source can we not just put a filter on the layer?

wnordmann:
Ok moved Filter to Layers and made a simple combine filter concept

examples/filter/app.js
```diff
@@ -0,0 +1,102 @@
...
+const store = createStore(combineReducers({
+  map: SdkMapReducer,
+}), window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
+   applyMiddleware(thunkMiddleware));
+
+function main() {
+  // Start with a reasonable global view of the map.
+  store.dispatch(mapActions.setView([-93, 45], 2));
+
+  store.dispatch(mapActions.setMapName('Basic Map Example'));
```
bartvde: not really needed, otherwise change the name

examples/filter/app.js
```diff
@@ -0,0 +1,102 @@
...
+  }));
+  const loadJson = (sourceName) => {
+    fetch('./states.geojson')
+      .then(
+        response => response.json(),
+        error => console.error('An error occured.', error),
+      )
+      // addFeatures with the features, source name
+      .then(json => {
+        store.dispatch(mapActions.addSource("states", {type:'geojson', data: json}));
```
bartvde: I would use ```data: './states/geojson'``` here to  make the example as simple as possible, no need for fetching in this example

examples/filter/app.js
```diff
@@ -0,0 +1,102 @@
...
+          },
+        }));
+      });
+  };
+  loadJson();
+
+  const reset = () => {
+    store.dispatch(mapActions.removeSource('states'));
+    store.dispatch(mapActions.removeLayer('states'));
+    loadJson();
```
bartvde: what is the purpose of the reset function? Can we not simply reset the filter on the layer?

examples/filter/app.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+  // place the map on the page.
+  ReactDOM.render(<SdkMap store={store} showZoomSlider />, document.getElementById('map'));
+
+  // add some buttons to demo some actions.
+  ReactDOM.render((
+    <div>
+      <FilterComponent store={store}/>
+      <button className="sdk-btn" onClick={reset}>Reset Map</button>
+    </div>
```
bartvde: since there are 2 reset buttons in the app, it might be confusing for users

examples/filter/filter.js
```diff
@@ -0,0 +1,131 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import * as mapActions from '@boundlessgeo/sdk/actions/map';
+
+// Custom Filter Component
+class filterComponent extends React.PureComponent{
+	constructor(props) {
+    super(props);
```
bartvde: check the indent in this file

examples/filter/filter.js
```diff
@@ -0,0 +1,131 @@
...
+		};
+
+    this.handleChange = this.handleChange.bind(this);
+  }
+	// Change handler to keep track of state
+  handleChange(event) {
+    this.setState({[event.target.name]: event.target.value});
+  }
+
+	// Action on save
```
bartvde: Save or Add Filter?

examples/filter/filter.js
```diff
@@ -0,0 +1,131 @@
...
+  addFilter() {
+		let value;
+		if(parseFloat(this.state.filterValue)){
```
bartvde: is this to check for numbers? Should we use isNaN here then?

examples/filter/filter.js
```diff
@@ -0,0 +1,131 @@
...
+			value = parseFloat(this.state.filterValue);
+		} else {
+			value = this.state.filterValue;
+		}
+		const layers = this.props.map.layers;
+		let currentFilter;
+		for (let i = 0, ii = layers.length; i < ii; i++) {
+			if (layers[i].id === "states") {
```
bartvde: we should probably make this configurable as a prop?

examples/filter/filter.js
```diff
@@ -0,0 +1,131 @@
...
+				}
+			}
+		}
+
+		this.props.updateLayer('states', {filter: newFilter});
+  }
+	// Action on reset
+  reset() {
+
+		const filter = [];
```
bartvde: any reason to use an empty array instead of null?

examples/filter/filter.js
```diff
@@ -0,0 +1,131 @@
...
+		this.props.updateLayer('states', {filter});
+  }
+
+	// Get list of keys
+  getKeyOptions() {
+		if( this.props.map.sources['states'] ){
```
bartvde: we should probably have a prop for the layer and read the source name from there?

examples/filter/filter.js
```diff
@@ -0,0 +1,131 @@
...
+		      }
+		    return options;
+			} else {
+				return [];
+			}
+		}
+  };
+
+  render() {
+		var options = this.getKeyOptions();
```
bartvde: is this something we would need to do each render, or can do it in componentDidMount e.g. ?

examples/filter/filter.js
```diff
@@ -0,0 +1,131 @@
...
+    return (
+	      <div className='modal-window'>
+					<div className='interior'>
+						<div className="bookmarkItem">
```
bartvde: is this className intentional? Or from the bookmarks example?

examples/filter/filter.js
```diff
@@ -0,0 +1,131 @@
...
+		         <button className="sdk-btn button-control" onClick={() => {this.addFilter()}} >Add Filter</button>
+		         <button className="sdk-btn button-control" onClick={() => {this.reset()}} >Reset</button>
+						</div>
+					</div>
+				</div>
+	    </div>
+	  )
+	}
+}
+// Getting the bookmark and map stores
```
bartvde: comment from other example?

examples/filter/filter.js
```diff
@@ -0,0 +1,131 @@
...
+function mapStateToProps(state) {
+  return {
+    map: state.map,
+    filter: state.filter,
```
bartvde: is there any filter in the state tree? Or is this a left-over?

examples/filter/filter.js
```diff
@@ -0,0 +1,131 @@
...
+  };
+}
+function mapDispatchToProps(dispatch) {
+  return {
+		removeFeatures: (sourceName, filter) => {
```
bartvde: not used anymore?

examples/filter/app.js
wnordmann: Add Source won't read the file do the ajax, unless I'm missing something.  I think it could be brought in with setContext, but I want to keep is explicit incase the user wanted to swap out the file.

bartvde: addSource will read it from url

examples/filter/app.js
wnordmann: No point.

examples/filter/app.js
wnordmann: Thought I removed it

examples/filter/app.js
bartvde: See data as a string here: https://www.mapbox.com/mapbox-gl-js/style-spec/#sources-geojson

@theduckylittle implemented this a while ago

examples/filter/filter.js
wnordmann: Good call much cleaner

examples/filter/filter.js
wnordmann: Null meets current implementation, it just feels like it is now a empty array

examples/filter/filter.js
wnordmann: I rewrote this a bit

examples/filter/filter.js
wnordmann: Started with bookmark, Adjusted the naming

bartvde:
I get a runtime warning when adding a filter:

```
warning.js:35 Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `filterComponent`. See https://fb.me/react-warning-keys for more information.
    in li (created by filterComponent)
    in filterComponent (created by Connect(filterComponent))
    in Connect(filterComponent)
    in div```

File: examples/filter/filter.js [hunk 16]
```diff
@@ -0,0 +1,152 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import * as mapActions from '@boundlessgeo/sdk/actions/map';
+
+// Custom Filter Component
+class filterComponent extends React.PureComponent{
+	constructor(props) {
+    super(props);
+		// List of keys to use from the loaded geojson to be filtered
+		const keysFromData = [
```
bartvde: if we want to do this, we should do it as a prop, and provide the config in app.js instead

examples/filter/filter.js
```diff
@@ -0,0 +1,152 @@
...
+			{name:"EMPLOYED", prettyName:"Number of people employed"},
+			{name:"UNEMPLOY", prettyName:"Number of people unemployed"},
+			{name:"P_MALE", prettyName:"Percent Male"},
+			{name:"P_FEMALE", prettyName:"Percent Female"}
+		]
+    this.state = {
+			operator:'==',
+			key:'',
+			filterValue: '',
+			keyOptions: keysFromData
```
bartvde: I don't think there is any need to have this in state, can be a prop instead

PR #652, opened 2017-09-12
update README.md to say build-examples rather than build:examples

qscripter:


bartvde:
@qscripter but the task name is https://github.com/boundlessgeo/sdk/blob/master/package.json#L14

theduckylittle:
bundle-examples is what's used for the static deploy and it actually does a lot of stuff that is distinct from build:examples (used for development).

PR #653, opened 2017-09-12
Added Bounding Box support to geojson type sources

theduckylittle:
1. Refactored the code to override the Vector Source's
   loader function. This allows SDK to interpret how features
   should be added to the source inside of the OpenLayers update
   cycle (specifically needed for BBOXes).
2. Added test to prove the URL will fetch with a bbox.

Refs: SDK-661


bartvde:
Reviewing this one now (after my first coffee ;-))

src/components/map.js
```diff
@@ -173,77 +175,80 @@ function configureMvtSource(glSource, accessToken) {
...
+
+function getLoaderFunction(mapProjection, baseUrl) {
+  return function(bbox, resolution, projection) {
+    const gl_source = this.get('_glSource');
+
+    // setup a feature promise to handle async loading
+    // of features.
+    let features_promise;
+
+    // if the data is a string, assume its a url
```
bartvde: it's

src/components/map.js
```diff
@@ -173,77 +175,80 @@ function configureMvtSource(glSource, accessToken) {
...
+      // relative to the style doc.
+      if (!(url.indexOf('http://') === 0 || url.indexOf('/') === '0')) {
+        if (baseUrl && url.indexOf('.') === 0) {
+          url = url.substring(1);
+        }
+        url = baseUrl + url;
       }
-      url = baseUrl + url;
+
+      // check to see if the bbox stratgy should be employed
```
bartvde: strategy

src/components/map.js
```diff
@@ -173,77 +175,80 @@ function configureMvtSource(glSource, accessToken) {
...
-    // if the baseUrl is present and the url does not
-    // start with http:// or "/" then assume the path is
-    // relative to the style doc.
-    if (!(url.indexOf('http://') === 0 || url.indexOf('/') === '0')) {
-      if (url.indexOf('.') === 0) {
-        url = url.substring(1);
+
+function getLoaderFunction(mapProjection, baseUrl) {
+  return function(bbox, resolution, projection) {
+    const gl_source = this.get('_glSource');
```
bartvde: I wonder if we can make getLoaderFunction an instance function instead so we can look up the source instead of tagging it onto the ol source object?

src/components/map.js
```diff
@@ -173,77 +175,80 @@ function configureMvtSource(glSource, accessToken) {
...
+        if (baseUrl && url.indexOf('.') === 0) {
+          url = url.substring(1);
+        }
+        url = baseUrl + url;
       }
-      url = baseUrl + url;
+
+      // check to see if the bbox stratgy should be employed
+      //  for this source.
+      const bbox_string = '{bbox-epsg-3857}';
```
bartvde: we should probably have this as a constant on a higher level and reuse it for WMS and WFS

src/components/map.js
```diff
@@ -173,77 +175,80 @@ function configureMvtSource(glSource, accessToken) {
...
+      if (url.indexOf(bbox_string) >= 0) {
+        url = url.replace(bbox_string, bbox.toString());
```
bartvde: since this is targeting WFS, maybe we should be explicit with our BBOXes and include the projection/SRS in it? https://gis.stackexchange.com/questions/225535/wfs-and-wms-query-srs-parameter-and-bbox-usage

src/components/map.js
```diff
@@ -254,8 +259,11 @@ function updateGeojsonSource(olSource, glSource, mapProjection, baseUrl) {
  *
  * @returns ol.source.vector instance.
  */
-function configureGeojsonSource(glSource, mapProjection, baseUrl, wrapX) {
+function configureGeojsonSource(glSource, mapView, baseUrl, wrapX) {
+  const use_bbox = (typeof glSource.data === 'string' && glSource.data.indexOf('bbox-epsg-3857') >= 0);
   const vector_src = new VectorSource({
+    strategy: use_bbox ? LoadingStrategy.bbox : LoadingStrategy.all,
+    loader: getLoaderFunction(mapView.getProjection(), baseUrl),
```
bartvde: alternatively, we could simply pass glSource here or not?

src/components/map.js
```diff
@@ -173,77 +175,80 @@ function configureMvtSource(glSource, accessToken) {
...
   if (glSource.cluster) {
-    vector_src = olSource.getSource();
-    if (glSource.clusterRadius !== olSource.getDistance()) {
-      olSource.setDistance(glSource.clusterRadius);
-    }
+    src = olSource.getSource();
   }
 
+  // update the glSource definition for the loader
+  src.set('_glSource', glSource);
```
bartvde: Ideally I'd really want to avoid this.

src/components/map.js
```diff
@@ -173,77 +175,80 @@ function configureMvtSource(glSource, accessToken) {
...
-        vector_src.addFeatures(GEOJSON_FORMAT.readFeatures(features, readFeatureOpt));
-      }
-    }).catch((error) => {
-      console.error(error);
-    });
-  }
+  src.clear(true);
+
+  // force a refresh
+  src.loadFeatures(mapView.calculateExtent(), mapView.getResolution(), mapView.getProjection());
```
bartvde: doesn't clear handle this already?

src/components/map.js
```diff
@@ -254,8 +259,11 @@ function updateGeojsonSource(olSource, glSource, mapProjection, baseUrl) {
  *
  * @returns ol.source.vector instance.
  */
-function configureGeojsonSource(glSource, mapProjection, baseUrl, wrapX) {
+function configureGeojsonSource(glSource, mapView, baseUrl, wrapX) {
+  const use_bbox = (typeof glSource.data === 'string' && glSource.data.indexOf('bbox-epsg-3857') >= 0);
```
bartvde: we should use a const for the ```bbox-epsg-3857```

src/components/map.js
theduckylittle: Okay, I removed the _glSource bits.

src/components/map.js
theduckylittle: Not when the feature set was already empty (IE first load)

src/components/map.js
theduckylittle: My concern was with the update cycle, so my initial idea of pushing the definition in privately was to handle cases when there was a change -- the new implementation should handle this appropriately as well, it just creates a new function with the new glStyle definition.

src/components/map.js
```diff
@@ -173,77 +176,79 @@ function configureMvtSource(glSource, accessToken) {
...
-    if (glSource.clusterRadius !== olSource.getDistance()) {
-      olSource.setDistance(glSource.clusterRadius);
-    }
+    src = olSource.getSource();
   }
 
+  src.set('loader', getLoaderFunction(glSource, mapView.getProjection, baseUrl));
+
+  // update the glSource definition for the loader
+  // src.set('_glSource', glSource);
```
bartvde: commented code can go

PR #655, opened 2017-09-13
Allow skipping examples for hosted build

bartvde:
SDK-669

Also moved the artefacts into ```build/hosted``` not to conflict with the artefacts from the dev server build.


webpack-example.config.js
```diff
@@ -0,0 +1,71 @@
...
+const { lstatSync, readdirSync } = require('fs')
+const { join } = require('path')
+
+const isDirectory = source => lstatSync(source).isDirectory()
+const getDirectories = source =>
+  readdirSync(source).map(name => join(source, name)).filter(isDirectory)
+
+const subDirs = getDirectories('./examples');
+
+const entry = {};
```
theduckylittle: Since this code is the same between all the configs, should we break it out into a separate JS file that has like a "getEntryPoints()" function? Just so we don't worry about "drift" between the two build files.  I'm going to approve anyway but it's just a thought.

bartvde:
I just noticed it's also filtering the examples on the dev server, which I didn't expect, so need to figure out why

bartvde:
ah right I see now why it's happening, need to have a switch

bartvde:
That issue is resolved now, I'll look into generating a file for the common code now.

PR #659, opened 2017-09-14
popup app w/ imgs

CarlyLanglois:
example app includes popups w/ images loaded from relative urls. also tested w/ absolute urls with same effect.

Known issue right now - the ol.Overlay autopan/ pan-into-view action is taking place before the img is loaded in the popup so is not changing the view to sufficiently view the whole popup w/ the img.
The ImagePopup component has an internal state change based on the img loaded status but not sure how to leverage this status to use it for the autopan issue


examples/image-popup/app.js
```diff
@@ -0,0 +1,164 @@
+/** Image Popup application.
+ *
+ *  Contains a Map and demonstrates displying a
```
bartvde: displaying

examples/image-popup/app.js
```diff
@@ -0,0 +1,164 @@
...
+import '@boundlessgeo/sdk/stylesheet/sdk.scss';
+
+/* eslint-disable no-underscore-dangle */
+const store = createStore(combineReducers({
+  map: SdkMapReducer,
+}), window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
+   applyMiddleware(thunkMiddleware));
+
+/** A popup for marking features when they
+ *  are selected.
```
bartvde: outdated comment

examples/image-popup/app.js
```diff
@@ -0,0 +1,164 @@
...
+    const feature_imgs = this.props.features.map(f => f.properties.image);
+    const feature_titles = this.props.features.map(f => f.properties.title);
+
+    return this.renderPopup((
+      <div className="sdk-popup-content">
+        <p>
+          Breed(s) from here:<br />
+        <code>{ feature_titles.join(', ') }</code>
+        <br />
+        <img src={feature_imgs} alt={feature_ids} />
```
bartvde: since we can only display a single image, maybe we should only pass a single feature as prop (always take the first one)?

examples/image-popup/app.js
```diff
@@ -0,0 +1,164 @@
...
+    source: 'osm',
+  }));
+
+  // Add a geojson source to the map.
+  store.dispatch(mapActions.addSource('dogs', {
+    type: 'geojson',
+    data: {},
+  }));
+
+  // add a layer for the random points
```
bartvde: outdated comment

examples/image-popup/app.js
```diff
@@ -0,0 +1,164 @@
...
+            }
+          }
+
+          if (features.length === 0) {
+            // no features, :( Let the user know nothing was there.
+            map.addPopup(<SdkPopup coordinate={xy} closeable><i>No dogs here.</i></SdkPopup>);
+          } else {
+            // Show the super advanced fun popup!
+            const img = new Image();
+            img.onload = () => {
```
bartvde: we should probably add some info / comments for the reader here why we need to preload the image

examples/image-popup/index.html
```diff
@@ -0,0 +1,13 @@
+---
+layout: example.html
+title: Image Popup Example
+shortdesc: Demonstrates a popup containing an image on the map.
+docs: >
+  <li>Adding a GeoJSON layer to the map</li>
+  <li>Defining Popup classes for the map</li>
+  <li>Conditional styling for GeoJSON layers</li>
```
bartvde: that is not in this example right?

examples/image-popup/index.html
```diff
@@ -0,0 +1,13 @@
...
+  <li>Rendering an image inside of a popup</li>
+resources:
+  - https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css
+---
+<div id="controls"/>
```
bartvde: this div is not used so we might as well remove it

CarlyLanglois:
closing this to add the app to sdk-examples repo

PR #660, opened 2017-09-15
Add realtime data example

bartvde:
SDK-674

thanks to Mapbox for the inspiration https://www.mapbox.com/mapbox-gl-js/example/live-geojson/


bartvde:
seems this example broke with the recent changes for the bbox loader, will investigate

bartvde:
ah there is a call left where the projection is passed in instead of the mapview:

```updateGeojsonSource(this.sources[src_name], next_src, map_proj, this.props.baseUrl);
```

bartvde:
All fixed, ready again for review

bartvde:

PR #663, opened 2017-09-18
Refactor of layer list to include it in core.

theduckylittle:
1. Move the layer list component out of the example.
2. Created a very minimal vesion of a layer list for core which
   has a very stripped DOM representation and no labels.
3. Rewrote the layer list example with the more complex version
   featuring additional controls.

Some findings:
1. The <Provider/> was necessary to ensure the store was passed
   to child elements of the layer list.
2. It was not possible to simply override the `renderLayer` method
   of a parent class because `connect()'ed classes wouldn't honour
   the override.
3. Less a finding and more question: moving the layer up and down
   could probably be moved to the reducer/actions if it is considered
   common enough.  Having the code in the layer list may also be enough
   to make the point moot.


src/components/layer-list.js
```diff
@@ -0,0 +1,121 @@
+/** SDK Example Layerlist Component
```
bartvde: we should probably remove the word Example here?

src/components/layer-list.js
```diff
@@ -0,0 +1,121 @@
...
+import { connect } from 'react-redux';
+import PropTypes from 'prop-types';
+
+import { isLayerVisible } from '../util';
+
+import * as mapActions from '../actions/map';
+
+export class LayerListItem extends React.Component {
+
+  getLayerIndexById(id) {
```
bartvde: this might be good for util as well? ```getLayerIndexById(layers, id)``` ?

examples/layer-list/app.js
```diff
@@ -14,13 +14,53 @@ import SdkMap from '@boundlessgeo/sdk/components/map';
...
 
 /* eslint-disable no-underscore-dangle */
 const store = createStore(combineReducers({
   map: SdkMapReducer,
 }), window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__());
 
+class LayerListItem extends SdkLayerListItem {
+  render() {
+    const layer = this.props.layer;
+    const checkbox = this.getVisibilityControl(layer);
```
bartvde: would there be a way where we would not have to do this in every render? But e.g. just once in componentDidMount?

src/components/layer-list.js
```diff
@@ -0,0 +1,121 @@
...
+        checked={is_checked}
+      />
+    );
+  }
+
+  render() {
+    const layer = this.props.layer;
+    const checkbox = this.getVisibilityControl(layer);
+    return (
+      <li className="layer" key={layer.id}>
```
bartvde: seems all our current classNames have sdk- as the prefix

src/components/layer-list.js
```diff
@@ -0,0 +1,121 @@
...
+        <span className="name">{layer.id}</span>
+      </li>
+    );
+  }
+}
+
+LayerListItem.PropTypes = {
+  layer: PropTypes.shape({
+    id: PropTypes.string,
+    source: PropTypes.string,
```
bartvde: is source needed for a correct functioning of the layer list? Seems only id is used right?

src/components/layer-list.js
```diff
@@ -0,0 +1,121 @@
...
+  toggleVisibility: PropTypes.func.isRequired,
+  removeLayer: PropTypes.func.isRequired,
+  layers: PropTypes.arrayOf(PropTypes.shape({
+    id: PropTypes.string,
+    source: PropTypes.string,
+  })).isRequired,
+};
+
+LayerList.defaultProps = {
+  layerClass: LayerListItem,
```
bartvde: layerClass should be in the propTypes list as well right?

src/components/layer-list.js
```diff
@@ -0,0 +1,121 @@
...
+    return (
+      <ul>
+        { layers }
+      </ul>
+    );
+  }
+}
+
+LayerList.propTypes = {
+  toggleVisibility: PropTypes.func.isRequired,
```
bartvde: not used anymore as prop?

src/components/layer-list.js
```diff
@@ -0,0 +1,121 @@
...
+  removeLayer: PropTypes.func.isRequired,
```
bartvde: not used anymore as prop?

src/components/layer-list.js
```diff
@@ -0,0 +1,121 @@
...
+  layers: PropTypes.arrayOf(PropTypes.shape({
+    id: PropTypes.string,
+    source: PropTypes.string,
```
bartvde: source is probably not needed for the shape definition

theduckylittle:
Yes, I wanted to make sure the approach was acceptable to everyone *before* I started writing tests.  I'll be integrating this feedback and writing tests next.

examples/layer-list/app.js
theduckylittle: I don't think so. The checkbox needs to re-render everytime the visibility changes.  We could make the checkbox a subcomponent that has an even more limited scope -- which would prevent needing to re-render that part of the DOM tree -- but I think that even with a couple of hundred layers the shadow-dom is unlikely to grow to unmanageable proportions.

bartvde: ok I was thinking we could handle the visibility with the state of the component instead, but not sure if we would be solving any real performance issue here, haven't benchmarked or anything

bartvde:
looks good now code-wise, with tests this is good to go IMHO

theduckylittle:
Tests are built and ready.  Test coverage should creep back to 98+% when finished.

__tests__/components/layer-list.test.js
```diff
@@ -0,0 +1,161 @@
...
+        <button className="btn-remove" onClick={() => { this.removeLayer(); }}></button>
+      </div>
+    );
+  }
+}
+
+describe('test the LayerList component', () => {
+  let store = null;
+
+  // this is the same setup as used in legends but intead
```
bartvde: instead

__tests__/components/layer-list.test.js
```diff
@@ -0,0 +1,161 @@
...
+  //  of listing layers this lists the layers in a list.
```
bartvde: do you mean listing legends?

__tests__/components/layer-list.test.js
```diff
@@ -0,0 +1,161 @@
...
+  it('should move a layer up', () => {
+    const layers = store.getState().map.layers;
+
+    const wrapper = getCustomLayerList();
+    wrapper.find('.btn-up').last().simulate('click');
+
+    let new_layers = store.getState().map.layers;
+    expect(new_layers[0].id).toBe(layers[1].id);
+
+    // try to move a layer up that's alreaday at the top
```
bartvde: already

__tests__/components/layer-list.test.js
```diff
@@ -0,0 +1,161 @@
...
+    const layers = store.getState().map.layers;
+    const n_layers = layers.length;
+
+    const wrapper = getCustomLayerList();
+    wrapper.find('.btn-down').first().simulate('click');
+
+    let new_layers = store.getState().map.layers;
+    expect(new_layers[n_layers - 1].id).toBe(layers[n_layers - 2].id);
+
+    // try to move a layer down that's alreaday at the bottom
```
bartvde: already

examples/layer-list/app.js
```diff
@@ -14,13 +14,53 @@ import SdkMap from '@boundlessgeo/sdk/components/map';
...
+        <button className="sdk-btn" onClick={() => { this.removeLayer(); }}>
+          { this.props.labels.remove }
+        </button>
+      </span>
+    );
+
+    return (
+      <li className="layer">
+        <span className="checkbox">{checkbox}</span>
+        <span className="name">{layer.id}</span>
```
bartvde: should we use the title for display if present?

```layer.metadata['bnd:title']```

PR #664, opened 2017-09-18
missed a typo/false test during geojson merge.

theduckylittle:
I added this test at the last minute to quell a curiosity I had,
the test passed but was technically incorrect.  This fixes that
bug.


bartvde:
looks good please merge

PR #666, opened 2017-09-19
Fix for the loader not updating properly with GeoJson layers

theduckylittle:
This more forcefully updates the loader using `loader_ =` instead of `set('loader',..` which didn't appear to actually work.

### DIFF
--- a/src/components/map.js
+++ b/src/components/map.js
@@ -236,7 +236,7 @@ function updateGeojsonSource(olSource, glSource, mapView, baseUrl) {
   }
 
   // update the loader function based on the glSource definition
-  src.set('loader', getLoaderFunction(glSource, mapView.getProjection, baseUrl));
+  src.loader_ = getLoaderFunction(glSource, mapView.getProjection(), baseUrl));
 
   // clear the layer WITHOUT dispatching remove events.
   src.clear(true);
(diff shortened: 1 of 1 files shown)


src/components/map.js
```diff
@@ -236,7 +236,7 @@ function updateGeojsonSource(olSource, glSource, mapView, baseUrl) {
   }
 
   // update the loader function based on the glSource definition
-  src.set('loader', getLoaderFunction(glSource, mapView.getProjection, baseUrl));
+  src.loader_ = getLoaderFunction(glSource, mapView.getProjection(), baseUrl);
```
bartvde: this is using a private openlayers property, kind of risky
Wonder if there is another way

bartvde: Maybe we should PR to openlayers for a setLoader API function, I can look into that

bartvde: See: https://github.com/openlayers/openlayers/pull/7274

theduckylittle: I know, I *hate* using the non-documented API.  When I did the original solution by copying in `glSource` using `set` it was to avoid this EXACT line of code.  But the `set('loader'...` was simply not working and it was a huge blocker. 

If it _feels_ cleaner, we could sub-class ol.source.Vector and have our own setLoader and loadFeatures methods which use our own `.loader` member instead of the `.loader_ =` hack.

bartvde: If my PR to ol gets accepted we can use setLoader, let's wait for that

bartvde: So for now we can leave it like it is

PR #667, opened 2017-09-19
Wrote "npm link" how-to

theduckylittle:
SDK needs to have the package "built" to properly work with
a project. This documents the steps to creating the package
and then linking to it for local use.


walker:
Should the gitignore contain package-lock.json or are you all going to be sporadically updating and locking the dependencies?

bartvde:
@walker I think we can add the lock file to gitignore for now, will make that change

PR #669, opened 2017-09-19
Drawbox

wnordmann:
Adding in ability to draw a box to meet needed for mobile

### DIFF
--- a/__tests__/components/map-drawing.test.js
+++ b/__tests__/components/map-drawing.test.js
@@ -115,18 +115,6 @@ if (HAS_CANVAS) {
       // if the drawing control has been added to the map then
       //  there should be 1 additional interaction.
       expect(ol_map.getInteractions().getLength()).toBe(n_interactions + 1);
-
-      // get the last interaction, which should be the drawing interaction
-      const draw = ol_map.getInteractions().item(n_interactions);
-
-      const dummy_feature = new Feature();
-      dummy_feature.setGeometry(new Point(5, 5));
-      draw.dispatchEvent({
-        type: 'drawend',
-        feature: dummy_feature,
-      });
-
-      store.dispatch(MapActions.setView([-45, -45], 11));
     });
 
     it('turns on a drawing tool for box', () => {
--- a/__tests__/components/map-drawing.test.js
+++ b/__tests__/components/map-drawing.test.js
@@ -159,6 +159,18 @@ if (HAS_CANVAS) {
       // if the drawing control has been added to the map then
       //  there should be 1 additional interaction.
       expect(ol_map.getInteractions().getLength()).toBe(n_interactions + 1);
+
+      // get the last interaction, which should be the drawing interaction
+      const draw = ol_map.getInteractions().item(n_interactions);
+
+      const dummy_feature = new Feature();
+      dummy_feature.setGeometry(new Polygon(5, 5));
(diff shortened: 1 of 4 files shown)


__tests__/components/map-drawing.test.js
```diff
@@ -129,6 +129,50 @@ if (HAS_CANVAS) {
...
+
+      // if the drawing control has been added to the map then
+      //  there should be 1 additional interaction.
+      expect(ol_map.getInteractions().getLength()).toBe(n_interactions + 1);
+
+      // get the last interaction, which should be the drawing interaction
+      const draw = ol_map.getInteractions().item(n_interactions);
+
+      const dummy_feature = new Feature();
+      dummy_feature.setGeometry(new Polygon(5, 5));
```
bartvde: this is not a valid Polygon, argument should at least be an array

__tests__/components/map-drawing.test.js
```diff
@@ -129,6 +129,50 @@ if (HAS_CANVAS) {
...
+      draw.dispatchEvent({
+        type: 'drawend',
+        feature: dummy_feature,
+      });
+
+      store.dispatch(MapActions.setView([-45, -45], 11));
```
bartvde: what is tested here actually? I don't see any expect anymore at the end.

src/components/map.js
```diff
@@ -1199,9 +1199,15 @@ export class Map extends React.Component {
 
       this.activeInteractions = [select];
     } else if (INTERACTIONS.drawing.includes(drawingProps.interaction)) {
-      const draw = new DrawInteraction({
-        type: drawingProps.interaction,
-      });
+      var drawObj = {type: drawingProps.interaction};
+      if (drawingProps.interaction === 'Box'){
```
bartvde: should use the constant here

src/components/map.js
```diff
@@ -1199,9 +1199,15 @@ export class Map extends React.Component {
 
       this.activeInteractions = [select];
     } else if (INTERACTIONS.drawing.includes(drawingProps.interaction)) {
-      const draw = new DrawInteraction({
-        type: drawingProps.interaction,
-      });
+      var drawObj = {type: drawingProps.interaction};
```
bartvde: I'd put this in the else statement (also let's use ```let```)

src/components/map.js
```diff
@@ -1199,9 +1199,15 @@ export class Map extends React.Component {
...
+      if (drawingProps.interaction === 'Box'){
+        const geometryFunction = DrawInteraction.createBox();
+        drawObj = {source: drawingProps.sourcename,
```
bartvde: source should not be set here, it's handled by the ```drawend``` event instead

src/components/map.js
```diff
@@ -1199,9 +1199,15 @@ export class Map extends React.Component {
...
+          type: 'Circle',
+          geometryFunction: geometryFunction
```
bartvde: can be simply ```geometryFunction,```

src/components/map.js
wnordmann: Totally, at first I thought you meant a const and I was confused

src/components/map.js
wnordmann: Easy enough to put in an else, but it's a really small else.

PR #674, opened 2017-09-21
added content to the landing page.

jjmulenex:
adjust the template and styles as necessary


examples/index.html
```diff
@@ -1,11 +1,66 @@
 ---
 layout: index.html
 title: Boundless SDK
-pageTitle: What is Boundless SDK?
+pageTitle: Welcome to Boundless SDK!
+skipExamplePage: true
```
bartvde: this is not needed anymore, we can filter out using the name of the template

bartvde: Removed with https://github.com/boundlessgeo/sdk/pull/686

PR #688, opened 2017-09-26
Added fontawesome/font-icon example

theduckylittle:
Uses the font-awesome icons to represent the
airports as little airplanes on the map.

It's both cute AND demonstrates using a font as an icon
alternative to a "circle" type layer.

### DIFF
--- a/examples/feature-table/app.js
+++ b/examples/feature-table/app.js
@@ -68,7 +68,6 @@ function main() {
       'text-field': '\uf072',
     },
     paint: {
-      //'text-color': '#552211',
       'text-color': '#CF5300',
     },
   }));
--- a/examples/feature-table/app.js
+++ b/examples/feature-table/app.js
@@ -61,7 +61,7 @@ function main() {
     type: 'symbol',
     source: 'dynamic-source',
     layout: {
-      'text-font': 'FontAwesome Regular',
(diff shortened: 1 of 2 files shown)


examples/feature-table/index.html
```diff
@@ -6,6 +6,8 @@
   <li>Adding an tiled map</li>
   <li>Adding a GeoJSON-type layer to a map using Fetch</li>
   <li>Displaying data in a simple table</li>
+  <li>Using Font Awesome to create map icons.</li>
 ---
+<script src="https://use.fontawesome.com/2f2c3f1eb2.js"></script>
```
bartvde: resources is meant for this in the template see https://github.com/boundlessgeo/sdk/blob/master/examples/export-image/index.html#L7:L8 for an example

examples/feature-table/app.js
```diff
@@ -58,12 +58,18 @@ function main() {
...
+      'text-size': 18,
+      'icon-optional': true,
+      // airplane icon
+      'text-field': '\uf072',
+    },
     paint: {
-      'circle-radius': 5,
-      'circle-color': '#552211',
-      'circle-stroke-color': '#00ff11',
+      //'text-color': '#552211',
```
bartvde: let's take out this comment

PR #689, opened 2017-09-27
Upgrade to React 16

bartvde:
Now that React 16 has been released, we can move to the latest and greatest


bartvde:
@theduckylittle confirmed he has no objections, so let's move to React 16

PR #690, opened 2017-09-28
Add style and className prop for all our components

bartvde:
SDK-701


bartvde:
This is now ready for review, TIA

PR #691, opened 2017-09-29
Fix for popup autoPan

bartvde:
Since #662 popup autoPan was broken. This PR adds a different solution than the DOM manipulation we did back then.

I'm not 100% happy with having to pass a reference to the map to the popup, but I couldn't think of a better approach.

@CarlyLanglois would be great if you can confirm that this fixes things for you.
@theduckylittle would like to have your feedback on this one

### DIFF
--- a/src/components/map.js
+++ b/src/components/map.js
@@ -895,6 +895,7 @@ export class Map extends React.Component {
     overlays.forEach((overlay) => {
       const id = overlay.get('popupId');
       if (this.popups[id].state.closed !== false) {
+        this.popups[id].setMap(null);
         // mark this for removal
         overlays_to_remove.push(overlay);
         // umount the component from the DOM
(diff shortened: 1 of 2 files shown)


bartvde:
For reference this came up in: https://github.com/boundlessgeo/sdk-examples/pull/6

theduckylittle:
This will be a constant chase and I think a few different iterations of fixes have all required hitting the DOM and/or ref'ing the DOM in various ways with varying degrees of ugly.  If this is fixing the example, I'm willing to give it a :+1: and move on.

CarlyLanglois:
yep - this works for me. thank you!

bartvde:
cool I'll rebase and merge then

thanks for the feedback @theduckylittle and @CarlyLanglois

PR #693, opened 2017-10-02
Have a generic controls prop instead of showZoomSlider

bartvde:
Instead of having specific props for controls, we can have a generic one.

We can leave creation of the controls up to the application, so we don't have a lot of potentially unused code in SDK core.


bartvde:
Closing this one, as we don't to encourage people to use controls

PR #694, opened 2017-10-04
Clean up package.json

bartvde:
This is an attempt at tackling SDK-689

Also up for discussion: should our peerDependency for React force 16+ or should we have both 15 and 16 to be more usable?


bartvde:
also @theduckylittle would it make sense to have ```canvas``` as an ```optionalDependency```?

theduckylittle:
Yes! Let's do that.

bartvde:
was just thinking about it a bit more and now doubting if it makes sense

since it's only for development (but there is no optionalDevDependency)

adding it to optionalDependency might mean it gets installed unnecessary when installing the sdk package

theduckylittle:
Right now, we have a problem with `ol` and any tests that involve it. Any test that includes `ol` will fail without `canvas`.  However, on Windows installing `canvas` is a heavy lift.  When I added checking for `canvas` it caught the majority of the issues.  But when we upgraded `ol` it broke more tests as `ol` now attempts to define a canvas element before it is really needed.  That breaks a lot of the tests that should otherwise function without `canvas`.

I will try to see how hard it is to get `canvas` working on my Windows box. If it's not as heavy of a lift as I first encountered then the point may be moot.

bartvde:
If it's devDependency, it should be okay IMHO
we don't recommend development on Windows anyway?

theduckylittle:
Let's add it as a devDependency then. Running tests on windows will be problematic but if someone is really dedicated then they can power through. ;)

bartvde:
I will add the change to this PR

bartvde:
thanks for the review, hope it doesn't end up breaking something :)
We will find out after we publish a new package I guess

PR #695, opened 2017-10-05
Create our own zoom and zoom slider controls

bartvde:
SDK-709


bartvde:
this will probably also need some love from @jjmulenex 
the scss will probably need to use variables for colors

selection_401

PR #696, opened 2017-10-06
Vector legend

bartvde:
SDK-696

### DIFF
--- a/examples/mb-bright/app.js
+++ b/examples/mb-bright/app.js
@@ -8,10 +8,12 @@ import thunkMiddleware from 'redux-thunk';
 import React from 'react';
 import ReactDOM from 'react-dom';
 
+import { connect } from 'react-redux';
 import { Provider } from 'react-redux';
 
 import SdkMap from '@boundlessgeo/sdk/components/map';
 import SdkZoomControl from '@boundlessgeo/sdk/components/map/zoom-control';
+import SdkLegend from '@boundlessgeo/sdk/components/legend';
 import SdkMapReducer from '@boundlessgeo/sdk/reducers/map';
 import SdkPrintReducer from '@boundlessgeo/sdk/reducers/print';
 import SdkMapboxReducer from '@boundlessgeo/sdk/reducers/mapbox';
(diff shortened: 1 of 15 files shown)


bartvde:
Will hold off on adding tests until this approach gets a basic review

theduckylittle:
Can we just give the Legend enough data in order to re-create the style function instead of passing the olLayer around?

bartvde:
We can but I thought it would be inefficient. So you mean just creating a dummy ol layer in the legend code and applying a new style function on there?

bartvde:
It would likely mean abstracting some of the code from the map component out into utlities? Given that we might also need to apply the sprites etc?

theduckylittle:
That's my thinking.  It would create an inefficiency but I think the impact will be relatively small because that code will only be run when the list of layers changes.

I feel like abstracting that code a bit could actually be beneficial.  We don't necessarily need to put it into `util` but could `export` the appropriate functions from map.

bartvde:
ok @theduckylittle I can work on those changes

bartvde:
If we want to add ```mapbox://``` sprite url support, we would need to pass apiKey and baseUrl also to the legend, or we would need some provider component that would provide that to all our components through context? @theduckylittle any advice on this specific question?

theduckylittle:
Should we add a MapBox reducer that tracks the mapbox-key? That way we could do "this.props.mapbox.apiKey" and check the auth against that? In such a way any component inside of an appropriate <Provider> would have access to the info necessary.

bartvde:
@theduckylittle that could also be an option, store it in the store
My initial concern was with serializing that content, but if we keep it separate from map it should be no issue

bartvde:
I will make that change in this PR

theduckylittle:
If it's in a separate reducer then it will not be serialized with the map.

bartvde:
Right, agreed.

examples/mb-bright/app.js
```diff
@@ -22,20 +24,20 @@ import CONFIG from './conf'; // eslint-disable-line
 const store = createStore(combineReducers({
   map: SdkMapReducer,
   print: SdkPrintReducer,
+  mapbox: SdkMapboxReducer,
 }), window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
    applyMiddleware(thunkMiddleware));
 
 function main() {
+  store.dispatch(mapboxActions.setBaseUrl('https://api.mapbox.com/styles/v1/mapbox/bright-v8'));
+  store.dispatch(mapboxActions.setAccessToken(CONFIG.access_token));
```
theduckylittle: Could we add: 
```
mapboxActions.configure({
 baseUrl:...,
 accessToken: ...,
});
```

The two actions will trigger two updates to the state tree.  And while that will be fast (no components are configured when the code runs) it would decrease the boilerplate for using MB sources.

bartvde: sure, so you mean just a single action right? And get rid of the two separate ones?

bartvde: @theduckylittle made the change with: https://github.com/boundlessgeo/sdk/pull/696/commits/ae7c6c3cb227b7beeede337de624ee542057911e

PR #697, opened 2017-10-06
Sort examples in index alphabetically

bartvde:
This has been bothering me for a while ;)


bartvde:
We should probably also add a search field at some point

PR #699, opened 2017-10-11
Configure WFS-T example for the demo geoserver

bartvde:
SDK-686


theduckylittle:
i'm merging the demo so that it works in the wild but someone on the demo server should consider simplifying the polygon.

PR #700, opened 2017-10-13
Canvas sprite animations

bartvde:
SDK-717


bartvde:
I think the functionality is all in place now, just needs the unit tests.

bartvde:
starting to work on unit tests, it seems the determineTainting method of ol.style.IconImage is "crashing" jest

src/components/map.js
```diff
@@ -647,20 +652,83 @@ export class Map extends React.Component {
...
+    for (i = 0, ii = layers.length; i < ii; ++i) {
+      const layer = layers[i];
+      spriteOptions[layer.id] = jsonClone(layer.metadata['bnd:animate-sprite']);
+      if (Array.isArray(layer.filter)) {
+        layer.filter = createFilter(layer.filter);
+      }
+    }
+    olLayer.setStyle((feature, resolution) => {
+      // loop over the layers to see which one matches
+      for (let i = 0, ii = layers.length; i < ii; ++i) {
```
theduckylittle: Does the linter throw up a warning for this line? `i` and `ii` are defined on line 663 and this could have side effects.

bartvde: the linter did not complain, but I will check it out

bartvde: I changed it, better to be safe than sorry ;-)

PR #702, opened 2017-10-17
Instructions on Ejecting Create-React-App

theduckylittle:
create-react-app will not be able to _build_ a bundle due to some issues with various SDK dependencies. These instructions provide detail on what to do next.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -229,7 +229,7 @@ Congratulations! You should have a fully operational Boundless SDK React app!
 
    ```
 
-5. Modify `package.json` by ading the following line after `"scripts": {`:
+5. Modify `package.json` by adding the following line after `"scripts": {`:
     ```
     "watch": "NODE_ENV=development webpack --watch --config ./config/webpack.config.dev.js",
     ```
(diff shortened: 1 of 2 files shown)


docs/getting-started.md
```diff
@@ -183,3 +183,32 @@ yarn start
...
+2. `yarn add webpack-cli`
+3. Modify `config/webpack.config.prod.js`, change 149 from:
+    ```
+    include: paths.appSrc,
+    ```
+    To:
+    ```
+    include: [paths.appSrc, '../node_modules'],
+    ```
+4. Modify `config/webpack.config.dev.js`, comment out line 46,
```
bartvde: is it safe to use line numbers? They might change in the future? So at least I would add the line here that needs commenting.

docs/getting-started.md
```diff
@@ -183,3 +183,32 @@ yarn start
...
+   hot reloading will not work inside of a served environment.
+
+5. Modify `package.json` by ading the following line after `"scripts": {`:
```
bartvde: adding

theduckylittle:
Updated instructions based on @bartvde feedback.

PR #705, opened 2017-10-25
Vector legend fixes

bartvde:
SDK-738

This PR fixes multiple issues with vector legends:

- legends will get overwritten
- legends did not use strokes
- legends from multiple layers were being the same as the last layer


bartvde:
@CarlyLanglois it would be great if you could confirm that this PR fixes the issues you were seeing with the vector legend in the styling app

CarlyLanglois:
yep! This fixes all issues. thanks

bartvde:
great, thanks for confirming @CarlyLanglois

PR #708, opened 2017-10-25
Two way edit612

wnordmann:
Adding in two way binding for feature table
Updating action/reducer to support adding feature at position
Moving tests for adding features to the map to own test file
Adding some styles for tables


examples/feature-table/table.js
```diff
@@ -0,0 +1,187 @@
+import React from 'react';
+// import ReactDOM from 'react-dom';
```
theduckylittle: this file has a collection of comments that should get removed.

examples/feature-table/table.js
```diff
@@ -0,0 +1,187 @@
...
+      Object.assign({}, this.state.editRecord, {[key]:value}) });
+  }
+}
+// Build the body of the table based on list of properties and source store in redux store
+buildTableBody (properties, sourceName) {
+  const body = [];
+  let row = [];
+  // Get all the features from the Redux store
+  if(sourceName === ''){
+    return;
```
theduckylittle: Returning undefined or '' does _work_ but react prefers returning false.

examples/feature-table/table.js
```diff
@@ -0,0 +1,187 @@
...
+  // Loop over features
+  for (let i = 0, ii = features.length; i < ii; i++) {
+    // Loop over properties
+    for (let j = 0, jj = properties.length; j < jj; j++) {
+      // Build list of properties for each feature
+      const featureValue = features[i].properties[properties[j]];
+      row.push(
+        <td key={j}>
+          {this.state.editRow === i || featureValue}
+          {this.state.editRow !== i || <input type="text" placeholder={featureValue} onBlur={(evt) => this.updateFeature(evt.target.value, properties[j])}/>}
```
theduckylittle: Can we make this more concise by always using an <input> but changing the value to read only if it's not editable?

src/actions/map.js
```diff
@@ -208,11 +208,15 @@ export function setClusterRadius(sourceName, radius) {
  *
  *  @returns {Object} Action object to pass to reducer.
  */
-export function addFeatures(sourceName, features) {
+export function addFeatures(sourceName, features, position) {
+  if (position === undefined){
```
theduckylittle: Instead of the if statement you can actually do:
```
export function addFeatures(sourceName, features, position = -1) {
```

src/actions/map.js
```diff
@@ -208,11 +208,15 @@ export function setClusterRadius(sourceName, radius) {
...
+    position = -1;
+  }
   return {
     type: MAP.ADD_FEATURES,
     sourceName,
     features,
+    position
```
theduckylittle: add the trailing comma here

src/reducers/map.js
```diff
@@ -293,10 +293,17 @@ function addFeatures(state, action) {
...
-      { features: data.features.concat(features) },
-    );
+    let featureCollection = data.features.concat(features);
+
+    if(action.position > -1){
+      // featureCollection = data.features.splice(action.position, 0, ...features);
+      featureCollection = [
+        ...data.features.slice(0, action.position),
+        ...features,
+        ...data.features.slice(action.position)
```
theduckylittle: trailing comma missing here

src/reducers/map.js
```diff
@@ -293,10 +293,17 @@ function addFeatures(state, action) {
...
     };
   } else if (data.type === 'FeatureCollection') {
-    new_data = Object.assign({},
-      data,
-      { features: data.features.concat(features) },
-    );
+    let featureCollection = data.features.concat(features);
+
+    if(action.position > -1){
+      // featureCollection = data.features.splice(action.position, 0, ...features);
```
theduckylittle: remove the dead code

examples/feature-table/table.js
```diff
@@ -0,0 +1,197 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      selectedSource: '',
+      editRow: -1,
+      editRecord: {}
+    };
+  }
+
+// Next few functions are all about building the feature Table
```
theduckylittle: The code from here on down got put on column 0 instead of column 2.

src/reducers/map.js
```diff
@@ -293,10 +293,16 @@ function addFeatures(state, action) {
       features: [data].concat(features),
     };
   } else if (data.type === 'FeatureCollection') {
-    new_data = Object.assign({},
-      data,
-      { features: data.features.concat(features) },
-    );
+    let featureCollection = data.features.concat(features);
```
theduckylittle: Because of how this will work, if action.position > -1 then this code will cause the array to be created like 3 times, completely spiking the memory when adding a feature.  Easy fix, have featureCollection = [] (or undefined) and put the default case as an `else` below.  Harder fix, use a for loop to construct a single new array. That prevents the creation of multiple (potentially large) arrays of features.

wnordmann: Good catch, thanks.  Wouldn't a better idea be to use use the same slice code just change the location of the slice to middle or end?

theduckylittle: I think based on this: https://jsperf.com/large-array-positioning/1

The fastest version of this would be to use:
```
let collection;
if (position > -1) {
  // do slice-and-concat
} else {
  // just concact
}
return collection;
```

PR #710, opened 2017-10-26
Empty vector legends should use empty text

bartvde:
SDK-739


bartvde:
this involves making getVectorLegend a method on the instance

CarlyLanglois:
looks great. Thanks for the time on this

PR #712, opened 2017-10-27
Handle exceptions in WFS-T

bartvde:
SDK-720

Tested with read-only GeoServer WFS and network error (shutdown Tomcat).


bartvde:
Ready for review

bartvde:

PR #714, opened 2017-10-31
Use pluggable maps

bartvde:
Now that ```ol``` has pluggable renderers (https://github.com/openlayers/openlayers/pull/6855) we should make use of them in sdk to get smaller builds


bartvde:
good point, I'll add a commit that does that

PR #715, opened 2017-10-31
Upgrade to webpack 3

bartvde:
SDK-705

I pegged most of the versions the same way they did in ```create-react-app``` https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/package.json


theduckylittle:
I was able to successfully upgrade using this branch and perform a build.

PR #716, opened 2017-11-01
Use different devtool to prevent CORS warning

bartvde:
This came up in SDK-666

It's to fix https://reactjs.org/docs/cross-origin-errors.html


theduckylittle:
Oh, snap, we're still using eval? Merging..

PR #718, opened 2017-11-02
Added better support for groups in the layer list

theduckylittle:
The layer list now accepts a class for rendering groups
as well as layers.

### DIFF
--- a/src/components/layer-list.js
+++ b/src/components/layer-list.js
@@ -101,10 +101,40 @@ SdkLayerListItem.PropTypes = {
   }).isRequired,
 };
 
+export class SdkLayerListGroup extends React.Component {
+
+  render() {
+    const children = [];
+
+    for (let i = 0, ii = this.props.layers.length; i < ii; i++) {
+        children.push(
+          <this.props.layerClass
+            exclusive={this.props.group.exclusive}
+            key={i}
+            layer={this.props.layers[i]}
+            groupId={this.props.groupId}
+          />
+        );
+    }
+
+    return (<li>{this.props.group.name}<ul>{children}</ul></li>);
+  }
+}
+
+SdkLayerListGroup.PropTypes = {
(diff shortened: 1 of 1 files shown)


src/components/layer-list.js
```diff
@@ -101,10 +101,40 @@ SdkLayerListItem.PropTypes = {
...
 
+export class SdkLayerListGroup extends React.Component {
+
+  render() {
+    const children = [];
+
+    for (let i = 0, ii = this.props.layers.length; i < ii; i++) {
+        children.push(
+          <this.props.layerClass
+            exclusive={this.props.group.exclusive}
```
bartvde: this prop is not documented in PropTypes

src/components/layer-list.js
```diff
@@ -101,10 +101,40 @@ SdkLayerListItem.PropTypes = {
...
+            key={i}
+            layer={this.props.layers[i]}
+            groupId={this.props.groupId}
```
bartvde: this prop is also not documented

src/components/layer-list.js
```diff
@@ -101,10 +101,40 @@ SdkLayerListItem.PropTypes = {
...
+          />
+        );
+    }
+
+    return (<li>{this.props.group.name}<ul>{children}</ul></li>);
+  }
+}
+
+SdkLayerListGroup.PropTypes = {
+  layerClass: PropTypes.func,
```
bartvde: this should be required right?

src/components/layer-list.js
```diff
@@ -157,6 +195,7 @@ SdkLayerList.propTypes = {
 
 SdkLayerList.defaultProps = {
   layerClass: SdkLayerListItem,
+  groupClass: SdkLayerListGroup,
```
bartvde: groupClass needs to be added to propTypes as well

PR #717, opened 2017-11-01
Layer list enhancements

bartvde:
When finished this PR will handle:

- hiding layers from the layer list (SDK-708)
- handle sorting of layers / groups in the layer list (SDK-744)
- handle influencing the HTML output (specifically no ```<ul>```, SDK-750)

PR #720, opened 2017-11-02
Layer list enhancements

bartvde:
follow-up PR from #717 

- hiding layers from the layer list (SDK-708)
- handle sorting of layers / groups in the layer list (SDK-744)


bartvde:

PR #723, opened 2017-11-06
Removed id="map" from template wrapper

walker:
This div id is dupe'd in the DOM by most of the examples.


walker:
I'm wrong. It's just dupe'd on filter example.

PR #726, opened 2017-11-07
Fix updateSource check for geojson

bartvde:
SDK-758


bartvde:
In the WFS-T example we were using updateSource with just a data key, so the geojson check failed. Instead we need to check for type on the mixin source of old and new.

bartvde:
it's slow because of the number of Mb going over the wire @theduckylittle has already filed SDK-723 to address this

PR #731, opened 2017-11-07
Fixing layer lit example to represent layer-list-item in a new file

wnordmann:
Quick adjustment to an import


CarlyLanglois:
this works for me but theres a huge extra polygon on the states layer and I cant tell where it's coming from - 

screen shot 2017-11-07 at 5 34 49 pm

wnordmann:
The state layer comes form https://demo.boundlessgeo.com I suspect it's a random edit.

bartvde:
we should consider letting travis build the examples so we catch this in PRs

bartvde:
yeah we had to enable WFS-T on the demo server for the WFS-T example which means anybody can change our data now ... wonder if that is what is happening

bartvde:
AFAICT there wasn't a way in geoserver to just enable WFS-T on a single layer

PR #732, opened 2017-11-08
Handle ordering of groups in layer list

bartvde:
SDK-766

One issue I found in the test cases was that there were cases where layer groups were not together in the stack, so I corrected those tests since you cannot have [layerA, groupA, layerB, groupA] situations, groupA needs to be together.


bartvde:
I'm finding now that placeLayer only seems to work in a certain sequence, so only if the source is the above the target, not the other way round, didn't expect that

PR #733, opened 2017-11-08
WMS and Identify support for vector tiles.

theduckylittle:
1. Added the ability to use vector tiles in identify.
   Vector tile layers will not return a fully hydrated feature,
   now the identify function picks up on that.
2. Also added BBOX style URL substitutions that allow using WMS
   to return a protobuf "image".

### DIFF
--- a/src/components/map.js
+++ b/src/components/map.js
@@ -47,7 +47,9 @@ import TileGrid from 'ol/tilegrid';
 
 import VectorTileLayer from 'ol/layer/vectortile';
 import VectorTileSource from 'ol/source/vectortile';
+
 import MvtFormat from 'ol/format/mvt';
+import RenderFeature from 'ol/render/feature';
 
 import ImageLayer from 'ol/layer/image';
 import ImageStaticSource from 'ol/source/imagestatic';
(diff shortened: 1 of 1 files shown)


src/components/map.js
```diff
@@ -212,12 +214,25 @@ function configureMvtSource(glSource, accessToken) {
   } else {
     urls = [url];
   }
+
+  // check to see if the url uses bounding box or Z,X,Y
+  let tile_url_fn = undefined;
```
bartvde: no need to initialize to undefined explicitly?

src/components/map.js
```diff
@@ -212,12 +214,25 @@ function configureMvtSource(glSource, accessToken) {
...
+  if (url.indexOf(BBOX_STRING) !== -1) {
+    tile_url_fn = function(urlTileCoord, pixelRatio, projection) {
+      let img_src = url.slice();
```
bartvde: no need to slice since replace will return a new string and not modify the original one?

src/components/map.js
theduckylittle: No, this is just a "thing" I do by habit. I removed it. It's a mental block I have ... leaving a variable with no specification.

theduckylittle:
WRT testing: I'll work on that in a separate PR. A test would be better but because configureMvtSource does not get exported we would need to implement rewire to truly test that something meaningful was happening.

PR #735, opened 2017-11-14
Add drag and drop to layerlist

bartvde:
SDK-757


bartvde:
Ready for review, see JIRA for a bit of background on some of the decisions in this PR

PR #737, opened 2017-11-15
Implemented group re-arrangement.

theduckylittle:
New MOVE_GROUP action will change the place of a group
with-in the layers list.


theduckylittle:
Looking at #735, this may or may not be needed as it handles the 'special case' of groups more seamlessly.

src/reducers/map.js
```diff
@@ -170,6 +170,89 @@ function orderLayer(state, action) {
...
+  // get the real index of the next spot for the group,
+  // if the placeAt index is mid group then place_at should
+  // be the index of the FIRST member of that group.
+  let place_start = place_at;
+  const place_group = getGroup(state.layers[place_at]);
+
+  if (place_group) {
+    // find the first index of the place group
+    let new_place = -1;
+    for (let i = 0, ii = n_layers; i < ii && new_place < 0; i++) {
```
bartvde: redefine of i

src/reducers/map.js
```diff
@@ -170,6 +170,89 @@ function orderLayer(state, action) {
...
+      if (getGroup(state.layers[i]) === place_group) {
+        new_place = i;
```
bartvde: personally I'd prefer a break here instead of the check in the for loop

src/reducers/map.js
```diff
@@ -170,6 +170,89 @@ function orderLayer(state, action) {
...
+  }
+
+
+  // build a new array for the layers.
+  let new_layers = [];
+
+  // have the group layers ready to concat.
+  const group_layers = state.layers.slice(group_start, group_end + 1);
+
+  for (let i = 0, ii = n_layers; i < ii; i++) {
```
bartvde: another redefine of i and ii

src/reducers/map.js
```diff
@@ -561,6 +644,9 @@ function setZoom(state, action) {
   return Object.assign({}, state, {zoom});
 }
 
+/** Move an entire group of layers in the stack
+ *
```
bartvde: is this comment in the correct place?

__tests__/reducers/map-groups.test.js
```diff
@@ -0,0 +1,84 @@
...
+    const layer_ids = [];
+    for (let i = 0, ii = layers.length; i < ii; i++) {
+      layer_ids.push(layers[i].id);
+    }
+
+    expect(layer_ids).toEqual(expectedOrder);
+  }
+
+  function createDefaultLayers() {
+    createLayer('A');
```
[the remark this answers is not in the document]
theduckylittle: I didn't do that to allow for different tests for different combinations of layers and groups.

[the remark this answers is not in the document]
bartvde: ok, so with future tests in my mind

src/reducers/map.js
```diff
@@ -170,6 +170,89 @@ function orderLayer(state, action) {
...
+  let group_end = null;
+  for (let i = 0, ii = n_layers; i < ii; i++) {
+    const group = getGroup(state.layers[i]);
+    if (group === action.group) {
+      if (group_start === null || i < group_start) {
+        group_start = i;
+      }
+      if (group_end === null || i > group_end) {
+        group_end = i;
+      }
```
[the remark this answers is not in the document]
theduckylittle: https://jsperf.com/math-min-max-vs-ternary-vs-if -- On my machine, using if()s are performing ~80% faster than calling out to Math.min.

PR #738, opened 2017-11-16
added getLayerSource

brambow:
Added getLayerSource - a utility function for getting the source object of a map layer.  It's a common operation that's needed to get the underlying data in order to pass to other tools (e.g., turf).

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -229,11 +229,11 @@ export function getLayerTitle(layer) {
 export function getLayerSource(layer, sources) {
   const sourceKeys = Object.keys(sources);
 
-  for(let i=0; i<sourceKeys.length;i++) {
-    if(sourceKeys[i] === layer.source){
+  for (let i = 0; i < sourceKeys.length; i++) {
+    if (sourceKeys[i] === layer.source) {
       return sources[sourceKeys[i]];
     } else {
       return null;
     }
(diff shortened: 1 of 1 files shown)


src/util.js
```diff
@@ -218,3 +218,22 @@ export function getLayerTitle(layer) {
...
+ * @returns {Object} The source object of the layer.
+ */
+export function getLayerSource(layer, sources) {
+  const sourceKeys = Object.keys(sources);
+
+  for (let i = 0; i < sourceKeys.length; i++) {
+    if (sourceKeys[i] === layer.source) {
+      return sources[sourceKeys[i]];
+    } else {
+      return null;
```
bartvde: this won't work right? If the first item is not the layer source, null will get returned even if the source is on position 2. Try something like this instead:

```
let source = null;
for (let i = 0, ii = sourceKeys.length; i < ii; i++) {
  if (sourceKeys[i] === layer.source) {
     source = sources[sourceKeys[i]];
     break;
  }
}
return source;
```

Also please add a unit test

bartvde: Actually we can simply do this or not?
```
return sources[layer.source]
```

and don't need a function at all right?

brambow:
Not needed.

PR #741, opened 2017-11-16
Add additional helpers for configuring sources

theduckylittle:
1. Small change to package.json to include the jest-cli.
   I believe in older version of Jest this is included but I was
   getting an error that tests could not be run without expressly
   including the cli package.
2. New actions which short-hand the creation of complex source
   definitions.


bartvde:
Looks good I will merge and add tests in an upcoming PR

PR #744, opened 2017-11-17
Make sure we set circle-stroke-width

bartvde:
The stroke of the circle in paint-change example was never visible, this is because the default value of circle-stroke-width is 0

Before this was not enforced in mapbox-to-ol-style so we did not notice it


bartvde:
Example change only so merging this one

PR #745, opened 2017-11-18
Added getLayersByGroup func, exposed more dnd API.

theduckylittle:
1. Needed to expose the isOver and DragPrevew API's of
   react-dnd to improve catalog interactions.
2. Added a getLayersByGroup function which is useful for
   collecting a set of style layers to be saved in GeoServer.


theduckylittle:
Test added, coverage saved. Merging!

PR #746, opened 2017-11-20
Fixes to be more in line with Mapbox GL JS / style spec

bartvde:
- layers need to have a type or ref (SDK-722)
- filter cannot be null (SDK-774)

Also fixes SDK-776 (overlay not visible in export image example)

### DIFF
--- a/__tests__/reducers/map.test.js
+++ b/__tests__/reducers/map.test.js
@@ -61,7 +61,6 @@ describe('map reducer', () => {
           metadata: {
             'bnd:title': title,
           },
-          filter: null,
         },
       ],
     });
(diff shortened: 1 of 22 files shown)


bartvde:
@theduckylittle hold off on the review, I'm going to check if type should really be required according to mapbox style spec. It's enforced by Mapbox GL though, but I wonder if it's correct.

theduckylittle:
I won't merge. The changes looked small enough to not bother me too much.

bartvde:
I'm going to remove the throw because it's not fully correct according to the spec IMHO I've raised the issue here: https://stackoverflow.com/questions/47397217/why-is-type-or-ref-required-for-layer-in-mapbox-gl-js

But I can add it to the examples in any case. Will update the PR. Thanks for the approval.

PR #748, opened 2017-11-21
Fix broken paint change example

bartvde:
#747 broke the paint change example, this PR fixes it


bartvde:
example change only so merging

PR #749, opened 2017-11-21
Do not check for null filter but falsey instead

bartvde:
Since the filter is not initially null anymore after #746 we needed to update the filter example


bartvde:
See also: https://stackoverflow.com/questions/47414045/why-can-layer-filter-not-be-null-or-undefined-when-using-mapbox-gl-style-setstyl

bartvde:
So it seems they allow null or undefined according to the spec, but it raises errors. Empty array also not allowed.

bartvde:
Only left in the commit that fixes the current broken example with openlayers

bartvde:
Example code only, so merging this small fix

PR #750, opened 2017-11-22
Add Mapbox GL renderer

bartvde:
Now ready for review

this decrease in coverage seems to be the maximum possible, given that we cannot test Mapbox GL JS under node.


src/components/mapboxgl.js
```diff
@@ -0,0 +1,524 @@
...
+        const layer_name = features[i].layer.id;
+        if (!features_by_layer[layer_name]) {
+          features_by_layer[layer_name] = [];
+        }
+        features_by_layer[layer_name] = features[i];
+      }
+      resolve([features_by_layer]);
+    });
+    // no xy and hms properties here
+    const coordinate = {
```
theduckylittle: is this better than just using `const coordinate = [lat, lon];` ?

theduckylittle:
this all works and I approve, recognizing the drop in coverage. My only concern is not fixable here which is that I wish we were able to  _choose_ which libraries are included based on the renderer. As is, we'll be pulling a lot of code down which may never be used.

src/components/mapboxgl.js
bartvde: I was attempting to keep the same structure as here: https://github.com/boundlessgeo/sdk/blob/master/src/components/map.js#L1248:L1252

bartvde:
Thanks @duck, but given that we have marked them as peerDependencies and all peerDependencies are optional, it shoudn't mean everything will get pulled in automatically? I.e. if someone writes an app that uses the openlayers map, they include ```src/components/map``` and they make sure they have the ```ol``` peerDependency? If they want to use the mapbox renderer, they include ```src/components/mapboxgl``` and include the mapbox-gl-js, draw and turf peerDeps?

src/components/mapboxgl.js
theduckylittle: Oh sure, that makes total sense then.

PR #752, opened 2017-11-27
When receiving context, set layer and source version metadata

bartvde:
This came up during SDK-780, it made the code easier and more efficient to find differences if we set the metadata for layers and sources version when we receive a new context.


bartvde:
sure I can add that test

bartvde:
@CarlyLanglois I know you ran into this before and I didn't really listen at that time ;-)

PR #756, opened 2017-12-05
Upgrade ol and ol-mapbox-style

bartvde:
SDK-769

The labels are a bit crowded though, is that to be expected @ahocevar ? Is it because the labels are in multiple layers, or is the new collission detection supported for that use case as well?

selection_428

### DIFF
--- a/__tests__/components/map.test.js
+++ b/__tests__/components/map.test.js
@@ -300,9 +300,6 @@ describe('Map component', () => {
     spyOn(layer, 'setSource');
     instance.updateLayerSource('mapbox');
     expect(layer.setSource).toHaveBeenCalled();
-    spyOn(instance, 'applyStyle');
-    instance.updateSpriteLayers({layers});
-    expect(instance.applyStyle).toHaveBeenCalled();
   });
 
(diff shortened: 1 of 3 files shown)


ahocevar:
OpenLayers decluttering is on the level of OpenLayers layer instances. You're also sacrificing performance by having the same source rendered by multiple layers. I'd recommend to use `apply` instead of `applyStyle`, which will optimize OpenLayers layer instance creation.

We could also add mapbox-style source and layer ids to the OpenLayers layer instances to make them easier to work with after creation.

bartvde:
cc-ing @theduckylittle since he might remember some of these early design decisions

bartvde:
@ahocevar was the previous labelgun based decluttering not on a per layer basis? Because I can't recall we saw this back then.

theduckylittle:
The reason for using applyStyle instead of style was so that we had finer control over individual layer changes. It allows us to do more fine-grained diffs between previous and new layer states.

bartvde:
@ahocevar I just confirmed, but we actually have a single vector tile layer since we group layers by source

PR #757, opened 2017-12-06
Do not unconditionally include ol/mapbox css

bartvde:
this moves ol-mapbox-style to be a peerDependency, so we can switch between renderers


bartvde:
thanks for the review and the testing @theduckylittle

Issue #760, opened 2017-12-09
not able to run npm install


kratosmat:
Hi all,
I am not able to install the sdk from source, I receive the following errors:

kratos@kratos-almaviva:~/Developments/AMA/sdk$ npm install
npm WARN deprecated babel-preset-es2015@6.24.1: 🙌  Thanks for using Babel: we recommend using babel-preset-env now: please read babeljs.io/env to update! 
npm WARN deprecated coffee-script@1.12.7: CoffeeScript on NPM has moved to "coffeescript" (no hyphen)

> canvas@1.6.7 install /home/kratos/Developments/AMA/sdk/node_modules/canvas
> node-gyp rebuild

gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
Package cairo was not found in the pkg-config search path.
Perhaps you should add the directory containing `cairo.pc'
to the PKG_CONFIG_PATH environment variable
No package 'cairo' found
gyp: Call to './util/has_lib.sh freetype' returned exit status 0 while in binding.gyp. while trying to load binding.gyp
gyp ERR! configure error 
gyp ERR! stack Error: `gyp` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onCpExit (/home/kratos/.nvm/versions/node/v9.2.1/lib/node_modules/npm/node_modules/node-gyp/lib/configure.js:336:16)
gyp ERR! stack     at ChildProcess.emit (events.js:159:13)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:209:12)
gyp ERR! System Linux 4.8.0-54-generic
gyp ERR! command "/home/kratos/.nvm/versions/node/v9.2.1/bin/node" "/home/kratos/.nvm/versions/node/v9.2.1/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /home/kratos/Developments/AMA/sdk/node_modules/canvas
gyp ERR! node -v v9.2.1
gyp ERR! node-gyp -v v3.6.2
gyp ERR! not ok 
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.1.3 (node_modules/fsevents):
[...]

PR #761, opened 2017-12-11
Use dotenv for Mapbox API key

bartvde:
SDK-795

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 node_modules/
 coverage
 npm-debug.log
-conf.js
+.env
 build
 dist
 docs/out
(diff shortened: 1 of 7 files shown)


bartvde:
This came up in #760

bartvde:
We currently don't build any of the examples that require a mapbox key but I will verify anyway

bartvde:
@theduckylittle I verified it will work in the example production builds if we ever need that.

selection_434

I'll add it to that webpack config as well here.

theduckylittle:
Cool! My hope was that we could get a key on the examples server to actually let this one work now! Thanks!

bartvde:
yes that is my hope also still ;-)

PR #762, opened 2017-12-12
Add support for ArcGIS REST Identify

bartvde:
This adds support for ArcGIS REST Identify so we can display popups on top of Esri services

### DIFF
--- a/__mocks__/fetch-jsonp.js
+++ b/__mocks__/fetch-jsonp.js
@@ -1,0 +1,8 @@
+// __mocks__/fetch-jsonp.js
+'use strict';
+
+const fetchjsonp = jest.genMockFromModule('fetch-jsonp');
+
+module.exports = function(url) {
+  return fetch(url);
+};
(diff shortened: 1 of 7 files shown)


bartvde:
Question I had during the implementation: should we introduce a metadata key ```bnd:query-type``` to make it more explicit, or would this be overkill since we probably only get WMS (which can be determined from the ol source class) and Esri ArcGIS REST (which we can determine from the generic ```bnd:query-endpoint```, although theoretically we could also probably determine the identify endpoint by some url sleuthing. Any feedback on this @theduckylittle ?

theduckylittle:
What about doing more specific URL subsitution in query-endpoint.  Where we provide the necessary parameters to do an identify query? E.G.: Image size, x,y, bbox, and projection? Something like this:

`http://dummy.com/mapping/wms?SERVICE=WMS&REQUEST=GetFeatureInfo&I={i}&J={j}&WIDTH={w}&HEIGHT={h}&BBOX={bbox-epsg-3857}&PROJ={projection}&LAYER=awesome`

That would let us manually specify the URL formulation for both the ESRI layers and custom WMS layers, and even arbitrary end-points.  In fact if we supported I, J (pixel space) and X, Y (projected space) we could support a whole flavor of identify-endpoints not yet imagined.

bartvde:
@theduckylittle sure that makes sense, do you mind if I do that in a follow-up PR?

theduckylittle:
Yep, let's merge this!

PR #763, opened 2017-12-13
Add scheme tms support for XYZ raster

bartvde:
SDK-796


bartvde:
This is now ready for review

PR #764, opened 2017-12-13
Fix intermittent Travis failure

bartvde:
SDK-798


bartvde:
I cleaned up the HAS_CANVAS stuff, since we can assume canvas is present for dev settings. Also it was only present in one file and otherwise it would need to get added to many.

Moved the CanvasPattern and CanvasGradient stuff to the environment setup, with slightly different logic

bartvde:
since this is test config only, I'm going to merge this one, so we can also see if the intermittent build failure is actually resolved, please ping in this ticket if you see it again and your branch is rebased on master

PR #765, opened 2017-12-13
Make sure we attribute OSM

bartvde:
SDK-797


bartvde:
Hmm maybe it's better to change our examples to explicitly have the attribution, so it will also work in mapbox-gl-js

PR #767, opened 2017-12-18
Add read support for Esri ArcGIS Rest Feature Services

bartvde:
SDK-800 SDK-806

### DIFF
--- a/src/components/esri.js
+++ b/src/components/esri.js
@@ -37,6 +37,10 @@ class EsriController extends Component {
     this.view = new View();
   }
 
+  componentDidMount() {
+    this.fetchData();
+  }
+
   shouldComponentUpdate(nextProps) {
     if (nextProps.map.center[0] !== this.props.map.center[0] || nextProps.map.center[1] !== this.props.map.center[1] || nextProps.map.zoom !== this.props.map.zoom || nextProps.sources !== this.props.sources) {
       this.fetchData();
--- a/src/components/esri.js
+++ b/src/components/esri.js
@@ -22,6 +22,7 @@ import EsriJsonFormat from 'ol/format/esrijson';
 import GeoJsonFormat from 'ol/format/geojson';
 import Proj from 'ol/proj';
 
+import {encodeQueryObject} from '../util';
 import {updateSource} from '../actions/map';
 
 /** @module components/esri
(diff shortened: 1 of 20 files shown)


bartvde:
oh wait let me try and see if this works in mapbox-gl-js as well

bartvde:
Ready again for review

src/components/esri.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+    // This should always return false to keep
+    // render() from being called.
+    return false;
+  }
+
+  fetchData() {
+    this.view.setCenter(Proj.fromLonLat(this.props.map.center));
+    this.view.setZoom(this.props.map.zoom + 1);
+    const extent = this.view.calculateExtent(this.props.map.metadata[MAP_SIZE_KEY]);
```
theduckylittle: Should we move the MAP_SIZE_KEY into the new "MapInfo" reducer which would also store the mouse-position?

src/components/esri.js
```diff
@@ -0,0 +1,95 @@
...
+    for (let key in this.props.sources) {
+      const source = this.props.sources[key];
+      const bbox = `{"xmin":${extent[0]},"ymin":${extent[1]},"xmax":${extent[2]},"ymax":${extent[3]},"spatialReference":{"wkid":102100}}`;
```
theduckylittle: Maybe this should be an object which is serialized with JSON.stringify?

src/components/esri.js
```diff
@@ -0,0 +1,95 @@
...
+      const url = `${source.onlineResource}${source.featureLayer}/query/?f=json&returnGeometry=true&spatialRel=esriSpatialRelIntersects&geometry=${bbox}&geometryType=esriGeometryEnvelope&inSR=102100&outFields=*&outSR=4326`;
```
theduckylittle: Instead of manually forming the query parameters we should use a standard encoder like the one in util.

src/components/esri.js
```diff
@@ -0,0 +1,95 @@
...
+      fetchJsonp(url, {timeout: this.props.timeout}).then(
+        response => response.json(),
+      ).then((json) => {
```
theduckylittle: This should have a catch() to ensure parse errors are caught.

src/components/mapboxgl.js
```diff
@@ -192,6 +192,12 @@ export class MapboxGL extends React.Component {
     this.layersVersion = getVersion(this.props.map, LAYER_VERSION_KEY);
     // when the map moves update the location in the state
     if (this.map) {
+      this.props.setSize([this.mapdiv.offsetWidth, this.mapdiv.offsetHeight]);
+
+      this.map.on('resize', () => {
```
theduckylittle: If we make a new reducer for map size, then we could also recalculate the bounds and put that data in there as well.

src/components/esri.js
bartvde: Ok, will do that

src/components/esri.js
bartvde: good point

src/components/esri.js
bartvde: will add

src/components/mapboxgl.js
bartvde: ok

bartvde:
I'm going to make the requested changes here first, and the continue on #769 which will need some changes after this gets in

src/components/mapboxgl.js
bartvde: Good suggestion, but I'm going to leave this to when we have a second component needing that info, if that is okay.

bartvde:
Changes made, I'm going to merge this one so I can continue the work on #769 which depends on this

PR #768, opened 2017-12-18
No need to import app.css, this is handled by template

bartvde:
This is a small fix to one of our examples. The example HTML template already handles the inclusion of app.css if it's present. https://github.com/boundlessgeo/sdk/blob/master/tasks/build-examples.js#L61:L66


bartvde:
Example only so merging this one

PR #774, opened 2017-12-22
Add scaleline control

bartvde:
SDK-828


bartvde:
For the default css positioning, I could not find a way to use bottom, maybe @jjmulenex or @theduckylittle can help me there?

src/util.js
```diff
@@ -31,6 +31,16 @@ export function jsonEquals(objectA, objectB) {
 }
 
 
+/** Gets the resolution for a zoom level in spherical mercator.
+ *
+ * @param {number} zoom The zoom level.
+ *
+ * @returns {number} The resolution for that zoom level.
+ */
+export function getResolutionForZoom(zoom) {
```
theduckylittle: Do we need the calculation if this is available from the view object?

bartvde:
@theduckylittle updated according to your review

PR #775, opened 2017-12-27
Set map extent in mapinfo reducer

bartvde:
Set the map bounds in WGS84 in the mapinfo reducer


bartvde:
This is now ready for review

PR #779, opened 2018-01-02
Add requestHeaders to the addWfsSource (geojson) options

blyefd:
Accepts `params` object in the `options` parameter of `addWfsSource()`.  The params will be added to the query string.
Accepts `requestHeaders` object in the `options` parameter of `addWfsSource()`.  The request headers will be set in the request to the server.
eg: 
```
SdkMapActions.addWfsSource(id, url, featureType, {
    params: {foo: 'bar', blah: 'blah'},
    requestHeaders: {"X-AUTH-TOKEN": 'abcdef', another: 'header'}
})
```

### DIFF
--- a/src/actions/map.js
+++ b/src/actions/map.js
@@ -518,7 +518,7 @@ export function addWmsSource(sourceId, serverUrl, layerName, options = {}) {
  * @returns {Object} Action to create a new source.
  */
 export function addWfsSource(sourceId, serverUrl, featureType, options = {}) {
-  const params = {
+  const params = Object.assign({
     'SERVICE': 'WFS',
     'VERSION': '1.1.0',
     // projection is always fixed for WFS sources as they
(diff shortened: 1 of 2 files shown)


bartvde:
One concern I have here is that mapbox gl might not accept another parameter on the source config, which is not part of the spec, since the validation will fail then

blyefd:
I removed the params.  Only the requestHeaders remain.

theduckylittle:
The real issue is the way in which the `requestHeaders`.  The Map reducer is intended to serialize and deserialize documents which comply to the MapBox GL Style Spec, as seen here: https://www.mapbox.com/mapbox-gl-js/style-spec/

Putting `requestHeaders` on the source would violate the style spec.  The way we handle this is by adding new keys to the `map.metadata` part of the document.

PR #781, opened 2018-01-03
Scaleline might not always be in controls div

bartvde:
In castling, the scaleline control is not in the controls div, so make sure our css selector is not so fussy


bartvde:
css change only so merging this one without review

PR #783, opened 2018-01-05
If legend depends on sprite, we need to wait for it to load

bartvde:
This came up with creating a legend for the castling app, some legends would not show up. This is because the sprite has not yet been loaded. If applyStyle finishes loading of the sprite image it will call changed on the vector layer, so we can use that hook.


bartvde:
now for the hard part, writing a unit test

bartvde:
somehow nock is not intercepting the request done by ol-mapbox-style

bartvde:
ah it must be because of: https://stackoverflow.com/questions/41706042/nock-is-intercepting-my-request-but-my-ajax-request-is-erroring-out

PR #785, opened 2018-01-06
Add failing test for mapbox style spec

milafrerichs:
## What does this PR do?
It adds a failing test to show an error I got trying to use a mapbox style json. 

I tried to use a mapbox specific style spec with mapbox-streets and the
mapbox satellite and it failed because of the urls in the source
definitions.

This PR is for you to test and discuss here and me trying to help fix it :)


bartvde:
Looks like we only have the protocol subsitution for vector currently, so for raster it needs to be added.

bartvde:
But I see it also uses the webp format: https://developers.google.com/speed/webp/

bartvde:
png also works as a fallback, so I guess we could only do that for now in SDK

PR #792, opened 2018-01-11
Add ability for separate stroke layer for vector legends

bartvde:
SDK-826

### DIFF
--- a/src/components/legend.js
+++ b/src/components/legend.js
@@ -22,7 +22,7 @@ import Point from 'ol/geom/point';
 import Feature from 'ol/feature';
 import VectorLayer from 'ol/layer/vector';
 import {applyStyle} from 'ol-mapbox-style';
-import {jsonClone, getLayerById, parseQueryString, encodeQueryObject} from '../util';
+import {jsonEquals, jsonClone, getLayerById, parseQueryString, encodeQueryObject} from '../util';
 import {getFakeStyle, hydrateLayer} from './map';
 
 /** @module components/legend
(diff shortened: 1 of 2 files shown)


src/components/legend.js
```diff
@@ -225,7 +225,21 @@ export class Legend extends React.Component {
...
+    let strokeEqual = true;
+    if (this.props.strokeId) {
+      const nextStrokeLayer = getLayerById(nextProps.layers, this.props.strokeId);
+      const strokeLayer = getLayerById(this.props.layers, this.props.strokeId);
+      strokeEqual = jsonEquals(strokeLayer, nextStrokeLayer);
+    }
+    return !layerEqual || !strokeEqual;
+  }
+
+  getVectorLegend(layer, layer_src, strokeLayer) {
```
theduckylittle: Should this pass in an array of layers instead of enumerating two of them?

bartvde: We can, if we always assume the second one is the stroke layer

bartvde: Actually, while implementing, the code would become more complex actually, so I'm not sure if it's a good idea

bartvde: I tried to implement that approach, but the code actually became more complex and longer by doing so

bartvde:
@theduckylittle I did restructure it a bit according to your comment see https://github.com/boundlessgeo/sdk/pull/792/commits/c41a0f4c3da64dda69efad9f91b037faa5c26a86

PR #794, opened 2018-01-12
Allow adding draw and edit styling options

milafrerichs:
## What does this PR od?
It allows you to change the style of editing, selecting and modifying. 

If you want to change the way the drawing styles look like you can
change them by using DRAWING actions to set the styles for editing,
modifying and selecting
I used Mapbox GL Styles to allow to use them for mapbox gl as well and
they will be converted to OL styles using the mapbox-to-ol-style package


bartvde:
thanks for this excellent contribution @milafrerichs

PR #796, opened 2018-01-15
[WIP] Add TIME parameter to vector tile requests

bartvde:
SDK-843

needs more work still

### DIFF
--- a/src/components/map.js
+++ b/src/components/map.js
@@ -570,7 +570,10 @@ export class Map extends React.Component {
                 params.TIME = time;
                 newUrls.push(urlParts[0] + '?' + encodeQueryObject(params));
               }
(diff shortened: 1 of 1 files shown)


bartvde:
There are still several issues:

- setUrls does not update immediately, only on subsequent calls
- the tileLoadFunction seems to be getting detached after a call to setUrls

bartvde:
The disadvantage of having to use the clear though is that the map will flash

bartvde:
Maybe we can filter the vector data from the pbf tile instead?

bartvde:
It seems the time attributes are not being sent over though

bartvde:
I'm closing this one, vector based filtering is really the way to go here

PR #797, opened 2018-01-16
Replace single time attribute with start and end

bartvde:
When we initially designed the time based support, we had a single metadata attribute key for time based datasets. However is it common to have a start and an end attribute, where the second one is optional. This is also how it works in Exchange. So  this PR changes that behaviour. The actual filtering will still take place at the application level.

### DIFF
--- a/src/components/map.js
+++ b/src/components/map.js
@@ -75,7 +75,7 @@ import LoadingStrategy from 'ol/loadingstrategy';
 
 import {updateLayer, setView, setBearing} from '../actions/map';
 import {setMapSize, setMousePosition, setMapExtent, setResolution, setProjection} from '../actions/mapinfo';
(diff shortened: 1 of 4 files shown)


bartvde:
I added another change, which is needed to show a time based vector tile layer initially filtered, instead of the full dataset

src/components/map.js
```diff
@@ -858,6 +858,10 @@ export class Map extends React.Component {
         this.applyStyle(layer, layers, sprite);
         return layer;
       case 'vector':
+        const time = getKey(this.props.map.metadata, TIME_KEY);
+        if (time && layers[0].metadata && layers[0].metadata[TIME_START_KEY] !== undefined) {
+          layers[0].filter = this.props.createLayerFilter(layers[0], time);
```
bartvde: I am doubting if this approach is sane, since I am changing state here. But using updateLayer it would cause another update?

PR #798, opened 2018-01-17
Measure improvements

bartvde:
This PR handles two issues:

- Do not handle feature info if active interaction (if you combined the measure tools with a map that has includeFeaturesOnClick set to true, feature info callbacks would be run and empty popups would show)
- Add an action to know when the measure geometry / feature is done


bartvde:
I need to add another commit to this PR, which deals with knowing when the measure geometry is finalized (drawend).

bartvde:
This is now ready for review

PR #800, opened 2018-01-19
Implement minZoom and maxZoom

### LINKED ISSUE #799: minZoom level for map?
This is a question about zoom levels on the map.  I know how to set the minzoom level on an open layers map by passing in minZoom and maxZoom to the view but its not immediatly obvious to me how to set it with boundless sdk.  I tried using the updateView and passing it in that way, as it is done in one of the examples, but it doesn't seem to stop the map from being able to zoom out.  Basically what it happening, when i zoom out, the map zooms so far out that the map gets really small and only takes up about 1/2 the webpage, and the world is shown three times. 
Is there a way to stop the zoom out from happening past a certain zoom level?
Thanks!

bartvde:
fixes #799


bartvde:
thinking about this a bit more, it probably makes more sense as a metadata key at the map level instead, so also the zoom slider can have the same info, as well as the mapbox gl map component, instead of having to duplicate props
```
map: {
  metadata: {
    'bnd:minzoom': 10,
    'bnd:maxzoom': 15,
  }
}
```

bartvde:
Rebased and ready for review

PR #803, opened 2018-01-23
Change mapbox gl draw to use a single instance

milafrerichs:
## What does this PR do?
Updates the support for mapbox draw. It was not working properly. 

### Reasons:
multiple instances of Draw do not share any information and would
destroy all features and elements. rather than creating a new instance
we simply change the mode of them.
the static mode which does not allow anything is the default mode

### DIFF
--- a/__tests__/components/mapboxgl.test.js
+++ b/__tests__/components/mapboxgl.test.js
@@ -15,7 +15,8 @@ configure({adapter: new Adapter()});
 
 const createMapDrawMock = () => {
   return {
-    changeMode: () => {}
+    changeMode: () => {},
+    getAll: () => {},
   };
 };
 const createMapMock = () => {
@@ -512,11 +513,12 @@ describe('MapboxGL component', () => {
     const map = wrapper.instance();
     // mock up our GL map
     map.map = createMapMock();
+    map.draw = createMapDrawMock();
     const on = () => {};
     map.map.on = on;
-    spyOn(map.map, 'addControl');
+    spyOn(map.draw, 'changeMode');
     map.updateInteraction({interaction: 'measure:LineString'});
-    expect(map.map.addControl).toHaveBeenCalled();
+    expect(map.draw.changeMode).toHaveBeenCalled();
   });
 
   it('should call setView', () => {
@@ -739,11 +741,11 @@ describe('MapboxGL component', () => {
     const map = wrapper.instance();
     // mock up our GL map
     map.map = createMapMock();
-    map.onDrawRender({
-      getAll: () => {
-        return {features: [{geometry: {}}]};
-      }
-    });
+    map.draw = createMapDrawMock();
+    map.draw.getAll = () => {
+      return {features: [{geometry: {}}]};
+    };
+    map.onDrawRender({});
     window.setTimeout(function() {
       expect(props.setMeasureGeometry).toHaveBeenCalled();
       done();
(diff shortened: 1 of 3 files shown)


__tests__/components/mapboxgl.test.js
```diff
@@ -25,6 +30,7 @@ const createMapMock = () => {
     setCenter: () => {},
     setBearing: () => {},
     setZoom: () => {},
+    addControl: () => {},
```
bartvde: if we add it here, there are other places where we can remove it. Also would be good then to add removeControl here as well. See e.g. https://github.com/milafrerichs/sdk/blob/36ae79377121455b6812ec95f2e41721da31edf1/__tests__/components/mapboxgl.test.js#L132:L135

src/components/mapboxgl.js
```diff
@@ -29,6 +29,7 @@ import {setMeasureFeature, clearMeasureFeature} from '../actions/drawing';
 import {LAYER_VERSION_KEY, SOURCE_VERSION_KEY} from '../constants';
 
 import 'mapbox-gl/dist/mapbox-gl.css';
+import StaticMode from '@mapbox/mapbox-gl-draw-static-mode';
```
bartvde: we need to add this as a peerDep and devDep then right?

src/components/mapboxgl.js
```diff
@@ -130,6 +134,13 @@ export class MapboxGL extends React.Component {
         if (this.props.map.metadata !== undefined &&
             this.props.map.metadata[version_key] !== nextProps.map.metadata[version_key] && this.map) {
           this.map.getSource(src_name).setData(nextProps.map.sources[src_name].data);
+          if (this.draw) {
+            if (this.map.getSource(src_name).type === 'geojson') {
```
bartvde: do we need to ask this to the map or can we just check this.props.map.sources / nextProps.map.sources instead?

src/components/mapboxgl.js
```diff
@@ -222,6 +233,13 @@ export class MapboxGL extends React.Component {
       if (this.props.initialPopups.length > 0) {
         this.map.on('load', this.onMapLoad);
       }
+      if (!this.draw) {
+        var modes = MapboxDraw.modes;
```
bartvde: const instead of var?

src/components/mapboxgl.js
```diff
@@ -290,15 +315,15 @@ export class MapboxGL extends React.Component {
     let defaultMode;
     if (INTERACTIONS.drawing.includes(drawingProps.interaction)) {
       defaultMode = this.getMode(drawingProps.interaction);
-      const drawOptions = {displayControlsDefault: false, defaultMode};
-      if (drawingProps.editStyle) {
-        drawOptions.styles = drawingProps.editStyle;
```
bartvde: was this intentionally removed?

milafrerichs: yes, the problem is that the styles can only be added once it is initialized and not later. so I need to find a way to have them before the `configureMap`

bartvde: ah ok, maybe we should leave a TODO in the code then so we don't forget to tackle this later?

bartvde: or are you planning to tackle this as part of this PR?

milafrerichs: no, in a separate PR. maybe we can add a ticket right away.

bartvde:
thanks for your contribution @milafrerichs

PR #804, opened 2018-01-23
Better handle selection and deselection

### LINKED ISSUE #802: "Deselect" features event?
This might be more of a use question rather than a code issue:

I'm building a map that has selection interactivity.  As I select features (holding shift to add to selection), I'm populating an array of selected features.  When I click the map, and a feature isn't present, the selection style disappears.  I'm lost on how to go about triggering an emptying of my selected features array when the # of selected features on the map is zero.

Is there a recommended way to go about this?

In the openlayers documentation I see that the Select interaction has a [getFeatures method](http://openlayers.org/en/latest/apidoc/ol.interaction.Select.html#getFeatures), which seems like it'd be better than my method of creating an array.  Is it possible to use this in the context of an SDK map?

bartvde:
Fixes #802

### DIFF
--- a/__tests__/components/map-drawing.test.js
+++ b/__tests__/components/map-drawing.test.js
@@ -258,10 +258,45 @@ describe('Map component with drawing', () => {
     select.getFeatures().push(features[0]);
     select.dispatchEvent({
       type: 'select',
+      selected: [features[0]],
     });
 
     // ensure onFeatureEvent was called.
     expect(sdk_map.onFeatureEvent).toHaveBeenCalled();
+  });
+
+  it('handles deselect', () => {
+    const props = {};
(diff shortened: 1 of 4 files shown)


bartvde:
Since deselection can happen on multiple features (click outside of a feature when multiple features were selected with shift click), it was changed to an array of features.
I was doubting if I should do the same for the draw and modified callbacks as well, and always pass a collection.
I was also doubting if it should onFeaturesSelected now instead of onFeatureSelected

bartvde:
For consistency reasons (and less lines of code) I decided to always pass a collection in the onFeature callback function
I also decided not to rename ```onFeatureSelected``` to ```onFeaturesSelected```

PR #805, opened 2018-01-24
Legend should be visible if layer is invisible

bartvde:
SDK-855

### DIFF
--- a/src/components/legend.js
+++ b/src/components/legend.js
@@ -11,6 +11,7 @@
  * under the License.
  */
 
+import omit from 'lodash.omit';
 import fetch from 'isomorphic-fetch';
 import PropTypes from 'prop-types';
 import React from 'react';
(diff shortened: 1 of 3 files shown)


src/components/legend.js
```diff
@@ -225,16 +226,27 @@ export class Legend extends React.Component {
     return (layer !== nextLayer);
   }
 
+  layersEqual(layer1, layer2) {
+    let paintEqual = true;
+    if (layer1 && layer1.paint && layer2 && layer2.paint) {
+      paintEqual = jsonEquals(layer1.paint, layer2.paint);
+    }
+    let layoutEqual = true;
+    if (layer1 && layer1.layout && layer2 && layer2.layout) {
```
theduckylittle: Could we skip adding the dependency if we did this instead?

```
const layout1 = Object.assign({}, layer1.layout);
const layout2 = Object.assign({}, layer2.layout);
delete layout1.visibility;
delete layout2.visibility;
layoutEqual = jsonEquals(layout1, layout2);
```

bartvde: Sure

bartvde:
@theduckylittle made the changes to not depend on lodash-omit

PR #806, opened 2018-01-24
Fix modifying a feature broke the selection

milafrerichs:
## What does this PR do?

the mode `direct_select` needs a `featureId` to work properly. 
but the featureId is the internal mapboxdraw feature id.

so I added an options parameter to inDrawModify to accept additional options to be passed to `changeMode`

### DIFF
--- a/src/components/mapboxgl.js
+++ b/src/components/mapboxgl.js
@@ -288,11 +288,10 @@ export class MapboxGL extends React.Component {
       draw.changeMode(mode);
     }, 0);
   }
-  onDrawModify(evt, drawingProps, draw, mode) {
+  onDrawModify(evt, drawingProps, draw, mode, options = {}) {
     this.onFeatureEvent('modified', drawingProps.sourceName, evt.features[0]);
     window.setTimeout(function() {
-      // allow to draw more features
-      draw.changeMode(mode, {feature: evt.features[0]});
+      draw.changeMode(mode, options);
     }, 0);
   }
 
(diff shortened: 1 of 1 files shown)


src/components/mapboxgl.js
```diff
@@ -322,7 +321,7 @@ export class MapboxGL extends React.Component {
     } else if (INTERACTIONS.modify === drawingProps.interaction || INTERACTIONS.select === drawingProps.interaction) {
       this.draw.changeMode('simple_select');
       this.map.on('draw.update', (evt) => {
-        this.onDrawModify(evt, drawingProps, this.draw, 'direct_select');
+        this.onDrawModify(evt, drawingProps, this.draw, 'direct_select', {featureId: evt.features[0].id});
```
bartvde: and you cannot simply do this in line 294? :

```
draw.changeMode(mode, {featureId: evt.features[0].id});
```

bartvde: in other words, I am failing to understand why we need to pass in options, when the option can be derived from the evt which is passed already or not?

milafrerichs: we probably could, but `changeMode` accepts different kind of options.
And `draw_line_string` for example accepts `featureId` and this could mess things up potentially.

https://github.com/mapbox/mapbox-gl-draw/blob/master/docs/API.md#changemodemode-string-options-object-draw

bartvde: ok cool

PR #808, opened 2018-01-25
Add ability to set modes for drawing

milafrerichs:
## What does this PR do?
It adds the ability to change the modes for every interaction. 
And change the mode that will be used after the interaction is finished. 

### Why this change?
it is possible with mapbox draw to have custom modes, this PR adds the
ability to assign them to the map and change them with the call to the
interactions.
there will be a mode for the interaction itself and one to switch to
after the interaction is done.
for example you want to draw new features, but after the feature is done
you don't want toi create another feature but rather change the newly
created one.
### Fixes
_Checks if the listener was added before and does not add them again_

this caused a bug in our system, that changing the mode often would cause the callback to be called multiple times. 
prevent that by checking if the listener was already added. it has to be made with `bind`  and at this place, because we need more info than just the event.

### DIFF
--- a/__tests__/components/mapboxgl.test.js
+++ b/__tests__/components/mapboxgl.test.js
@@ -790,7 +790,7 @@ describe('MapboxGL component', () => {
       interaction: 'Modify',
       sourceName: 'geojson',
     };
-    map.updateInteraction(drawingProps)
+    map.updateInteraction(drawingProps);
     expect(map.currentMode).toEqual('simple_select');
     expect(map.afterMode).toEqual('direct_select');
   });
--- a/__tests__/components/mapboxgl.test.js
+++ b/__tests__/components/mapboxgl.test.js
@@ -752,6 +752,49 @@ describe('MapboxGL component', () => {
     }, 0);
   });
 
+  it('optionsForMode returns featureId object', () => {
+    const wrapper = shallow(<MapboxGL />);
+    const map = wrapper.instance();
+    // mock up our GL map
+    map.map = createMapMock();
+    map.draw = createMapDrawMock();
+    expect(map.optionsForMode('direct_select', {features: [{id: 1}]})).toEqual({featureId: 1});
+  });
+
+  it('setMode returns the currentMode if afterMode is not set', () => {
+    const wrapper = shallow(<MapboxGL />);
+    const map = wrapper.instance();
+    // mock up our GL map
+    map.map = createMapMock();
+    map.draw = createMapDrawMock();
+    expect(map.setMode('direct_select')).toEqual('direct_select');
+  });
+
+  it('setMode returns after if afterMode is set', () => {
+    const wrapper = shallow(<MapboxGL />);
+    const map = wrapper.instance();
+    // mock up our GL map
+    map.map = createMapMock();
+    map.draw = createMapDrawMock();
+    expect(map.setMode('direct_select', 'simple_select')).toEqual('simple_select');
+  });
+
+  it('setMode returns after if afterMode is set', () => {
+    const wrapper = shallow(<MapboxGL />);
+    const map = wrapper.instance();
+    // mock up our GL map
+    map.map = createMapMock();
+    map.draw = createMapDrawMock();
+    map.addedDrawListener = true;
+    const drawingProps = {
+      interaction: 'Modify',
+      sourceName: 'geojson',
+    };
+    map.updateInteraction(drawingProps)
+    expect(map.currentMode).toEqual('simple_select');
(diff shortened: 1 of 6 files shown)


bartvde:
alternatively, can we just set a flag when we have registered it and not set it again?

src/components/mapboxgl.js
```diff
@@ -311,32 +330,43 @@ export class MapboxGL extends React.Component {
...
-      });
+      this.currentMode = this.setMode(this.getMode(drawingProps.interaction), drawingProps.currentMode);
+      this.afterMode = this.setMode(this.currentMode, drawingProps.afterMode);
+      this.draw.changeMode(this.currentMode);
     } else if (INTERACTIONS.modify === drawingProps.interaction || INTERACTIONS.select === drawingProps.interaction) {
-      this.draw.changeMode('simple_select');
-      this.map.on('draw.update', (evt) => {
-        this.onDrawModify(evt, drawingProps, this.draw, 'direct_select', {featureId: evt.features[0].id});
-      });
+      this.currentMode = this.setMode('simple_select', drawingProps.currentMode);
```
bartvde: can we make all these strings consts at the top of the file?

So ```const SIMPLE_SELECT = 'simple_select';``` etc?

bartvde: is it necessary to store the currentMode on the instance? Glancing at the code I don't see the need for making it a member of the instance?

bartvde: but maybe you did it for consistency because for afterMode the need is there

src/components/mapboxgl.js
```diff
@@ -311,32 +330,43 @@ export class MapboxGL extends React.Component {
...
       // but are prefixed with "measure:"
       const measureType = drawingProps.interaction.split(':')[1];
-      defaultMode = this.getMode(measureType);
-      this.draw.changeMode(defaultMode);
-      this.map.on('draw.render', (evt) => {
-        this.onDrawRender(evt);
-      });
+      this.currentMode = this.setMode(this.getMode(measureType), drawingProps.currentMode);
+      this.draw.changeMode(this.currentMode);
+      if (!this.addedMeasurementListener) {
```
bartvde: should we initialize it to false since we do the same for addedDrawListener ?

bartvde: we should also set it to true here right?

src/components/mapboxgl.js
```diff
@@ -311,32 +330,43 @@ export class MapboxGL extends React.Component {
...
+    if (!this.addedDrawListener) {
+      if (drawingProps.sourceName) {
+        this.addedDrawListener = true;
+        const drawCreate = (evt) => {
+          this.onDrawCreate(evt, drawingProps, this.draw, this.afterMode, this.optionsForMode(this.afterMode, evt));
+        };
+        const drawModify =  (evt) => {
+          this.onDrawModify(evt, drawingProps, this.draw, this.afterMode, this.optionsForMode(this.afterMode, evt));
+        };
+        this.map.on('draw.create', drawCreate.bind(this));
```
bartvde: since we use arrow functions do we need to bind at all?

src/components/mapboxgl.js
```diff
@@ -469,6 +499,8 @@ MapboxGL.propTypes = {
   }),
   /** Initial popups to display in the map. */
   initialPopups: PropTypes.arrayOf(PropTypes.object),
+  /** Inital drawing modes that are available for drawing */
```
bartvde: Initial

src/components/mapboxgl.js
milafrerichs: yes, it is for consistency. but thanks for checking.

src/components/mapboxgl.js
milafrerichs: yes, you're right.

src/components/mapboxgl.js
milafrerichs: will try it out. but you're probably right.

__tests__/components/mapboxgl.test.js
```diff
@@ -752,6 +752,49 @@ describe('MapboxGL component', () => {
...
+  it('setMode returns after if afterMode is set', () => {
+    const wrapper = shallow(<MapboxGL />);
+    const map = wrapper.instance();
+    // mock up our GL map
+    map.map = createMapMock();
+    map.draw = createMapDrawMock();
+    expect(map.setMode('direct_select', 'simple_select')).toEqual('simple_select');
+  });
+
+  it('setMode returns after if afterMode is set', () => {
```
bartvde: title of this test is a copy from the previous one

PR #809, opened 2018-01-31
Async creation of sources

bartvde:
SDK-862

This came up when porting https://github.com/boundlessgeo/ol-mapbox-style/pull/39 to SDK.

So ```type: 'vector'``` comes in two forms:

1. url, this is the url to a TileJSON resource, which is then fetched and parsed, and only then can we construct the actual source
2. tiles, a set of tiles, this is the way we were using url before

So we were incorrectly using url before, where we should have been using tiles. And when using url, it will need async / promise based creation of sources. This had quite a bit of influence on the unit tests, where some of them now need to run in a setTimeout callback.

### DIFF
--- a/__tests__/actions/map.test.js
+++ b/__tests__/actions/map.test.js
@@ -505,7 +505,7 @@ describe('async actions', () => {
     const sourceName = 'my-source';
     const sourceDef = {
       type: 'vector',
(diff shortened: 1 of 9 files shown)


bartvde:
Just noticed in castling that layer visibility changes don't work anymore with this change, need to investigate

bartvde:
The issue is here: https://github.com/boundlessgeo/sdk/blob/master/src/components/map.js#L995:L997

applyStyle is not getting run because the jsons are equal somehow, both have visibility none

PR #814, opened 2018-02-06
Fix layer order issue

bartvde:
SDK-871


wnordmann:
Looks good, nice simple fix and a test.

PR #816, opened 2018-02-12
Do not define functions in loop

bartvde:
This fixes the latest 3 issues reported by SonarQube

selection_513

We should find a way to detect them earlier.


bartvde:
non-functional change, so merging without review

PR #818, opened 2018-02-12
moving side effects to componentWillReceiveProps

manuelbieh:
Moved side effects from `shouldComponentUpdate` to `componentWillReceiveProps` since side effects in `shouldComponentUpdate` should be avoided.


bartvde:
thanks, I will check this out tomorrow morning

manuelbieh:
Glad to hear that 😉

PR #819, opened 2018-02-13
Fix several issues with the drawing tools

bartvde:
SDK-880

1. no box zoom anymore in mapbox gl map
2. onFeatureXXX callbacks need to pass a collection now
3. sourceName in callback functions should be correct
4. lines are not visible when drawing with the openlayers map in the draw example

So the draw control disabled the box zoom, see: https://github.com/mapbox/mapbox-gl-draw/blob/master/src/setup.js#L36

So we should only create the draw control when we actually need it.

In addition in #804 we had changed the interface of the onFeatureXXX callback functions to pass a collection instead of a single feature, but this had not yet been changed in the mapbox gl map component, so the drawing example was broken for the mapbox renderer, this PR also addresses this

### DIFF
--- a/__tests__/components/mapboxgl.test.js
+++ b/__tests__/components/mapboxgl.test.js
@@ -107,6 +107,8 @@ describe('MapboxGL component', () => {
     const map = wrapper.instance();
     // mock up our GL map
     map.map = createMapMock();
+    map.map.on = () => {};
+    map.map.off = () => {};
     map.draw = createMapDrawMock();
     spyOn(map.map, 'setStyle');
     spyOn(map.map, 'setCenter');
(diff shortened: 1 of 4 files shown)


bartvde:
Seeing an additional issue, since we only register the listeners once, we don't get the updated sourceName anymore if they change through the props, so better to remove and re-added I think like @milafrerichs had originally suggested but my review changed that, will fix that in here as well

bartvde:
Okay, I can see another issue in the draw example, with the openlayers renderer, when drawing lines, they are only visible on drawend, will investigate that one as well here

bartvde:
Ah there is only an edit style defined in the draw example for polygons it seems

bartvde:
All issues addressed, now ready for review

bartvde:
thanks for the review @milafrerichs

PR #823, opened 2018-02-16
[WIP] Add prop to have different featureClass for MVT

bartvde:
Needed for SDK-874

by having a real Feature as the MVT feature class, we can get to the geometry which we need for highlighting


bartvde:
Closing this one, since in hindsight this will not be the solution, due to geometries being clipped at the tile boundaries

PR #824, opened 2018-02-16
Check for existence of sources before using them

manuelbieh:
Same problem as described here earlier this week: https://github.com/boundlessgeo/sdk/pull/817

Under certain (asynchronous) conditions it might happen that you try to add, change or remove a feature on a source which has not been added yet or has already been removed. With this PR functions which are accessing a specific source simply return the old state if the source does not yet or no longer exists.

### DIFF
--- a/src/reducers/map.js
+++ b/src/reducers/map.js
@@ -406,6 +406,9 @@ function removeSource(state, action) {
  */
 function changeData(state, sourceName, data) {
   const source = state.sources[sourceName];
+  if (!source) {
+    return;
+  }
   const src_mixin = {};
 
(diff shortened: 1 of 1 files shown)


bartvde:
code looks good, are you able to squash the 3 commits into 1?
Also, it might be best to work from a feature branch instead of master for PRs

manuelbieh:
Created a bugfix branch and re-opened the PR. Since I'm working on a fork I think it does not really matter if I try to merge a remote bugfix branch into master or my remote master, does it?

See https://github.com/boundlessgeo/sdk/pull/825

PR #825, opened 2018-02-16
Checking if source exists before accessing it

manuelbieh:
See https://github.com/boundlessgeo/sdk/pull/824


bartvde:
thanks @manuelbieh

PR #827, opened 2018-02-19
Add ability to get async feature info for vector tiles

bartvde:
SDK-874

In addition to the above, this PR also adds the option to clear a geojson source when new features are added.


theduckylittle:
I see the utility of having  the `clear` option for the `addFeatures` action but I think it might be less clear in the future.  The `removeFeautres` action should be able to remove all features with an undefined filter.  Otherwise, adding a `setFeatures` function (which essentially does `addFeatures` with the `clear` option) would likely be more descriptive of what's happening.  

So I see two options:
1. Just call `removeFeatures` with no filter.
2. Add a new action called `setFeatures`.

bartvde:
@theduckylittle I removed the clear option in the latest commit, will adapt the castling PR to call removeFeatures first

PR #845, opened 2018-03-09
Add ability to request a map redraw

theduckylittle:
1. Updated the MapInfo reducer to handle a 'redraw request'.
   The 'MAPINFO.REQUEST_REDRAW' action increments the requestedRedraws,
   and then the map compares previous and current requestedRedraws values.
   When the current and next state's requestedRedraws is different then the
   map will trigger a redraw.
2. util has a new function `optionalEquals` which checks if a specific
   subtree of the state exists and then compares the values across the trees.
   This is a "safe" wrapper for optional reducers (e.g. MapInfo).
3. The mapboxgl component has been updated to honour resize and redraw requests.

Refs: SDK-915


src/components/mapboxgl.js
```diff
@@ -158,6 +158,12 @@ export class MapboxGL extends React.Component {
         }
       }
     }
+    // trigger a resie event when the size has changed or a redraw is requested.
```
bartvde: resize

src/util.js
```diff
@@ -278,3 +278,26 @@ export function getLayersByGroup(layers, groupId) {
   }
   return group_layers;
 }
+
+/** Compare the value of a prop from an optional state-tree.
+ *
+ *  @param {Object} state - Current state.
+ *  @param {Object} nextState - Next state.
+ *  @param {Object} reducerName - The "subtree" that may or may not exist.
```
bartvde: string instead of object?

theduckylittle:
Added [WIP] status as the updateSize call will redraw/render but doesn't force the layers to re-request data.

bartvde:
Ah right was just wondering if resize would do the complete trick or not

src/components/map.js
```diff
@@ -652,12 +704,31 @@ export class Map extends React.Component {
...
+          src.updateParams({'_ck': timestamp});
+        } else {
+          src.set('_ck', timestamp);
+          src.refresh({force: true});
+        }
+      }
+      // ensure the redraw happens.
+      // this.map.updateSize();
+    }
+    */
```
bartvde: this can go?

theduckylittle:
Removed.

PR #846, opened 2018-03-09
Opacity control for rasters

theduckylittle:
1. Adds styling for raster opacity.
2. Adds a demo of the raster-opacity to the basic demo.

refs: SDK-906


examples/basic/app.js
```diff
@@ -167,6 +171,43 @@ function main() {
...
+            defaultValue={opacity}
+            ref={(me) => {
+              this.opacityInput = me;
+            }}
+          />
+
+          <button
+            className="sdk-btn"
+            onClick={() => {
+              const v = this.opacityInput.value / 100;
```
wnordmann: `v` is a vague variable name

src/components/map.js
```diff
@@ -860,6 +860,12 @@ export class Map extends React.Component {
       applyStyle(olLayer, fake_style, layers[0].source);
     }
 
+    if (layers.length === 1 && layers[0].type === 'raster') {
```
wnordmann: This line is inside a forloop, if` layers.length === 1` then `layers[0]  === layer`, use layer instead

theduckylittle: This isn't inside a for loop. The for-loop ends on line 848.  With the way raster layers are entered, they only have one layer per source, so the loop isn't necessary.

theduckylittle:
@wnordmann - Updated!

PR #850, opened 2018-03-13
Move to new SonarQube instance

bartvde:
Also add the PR integration, and have a dependency check in a separate stage


bartvde:
CI change only so merging

PR #852, opened 2018-03-15
Psbex 119

travislbrundage:
Fixes wrong logic for proxying remote services in GeoNode client.

### DIFF
--- a/src/services/MapConfigService.js
+++ b/src/services/MapConfigService.js
@@ -60,24 +60,22 @@ class MapConfigService {
       return source;
     }
     var sourceObj = new ol.source[config.type](props);
-    
-    // Always proxy if config's use_proxy = True
-    // otherwise use original conditional
-    if (opt_proxy && typeof config['use_proxy'] !== "undefined" && config['use_proxy'] === true) {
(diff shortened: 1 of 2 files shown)


src/services/MapConfigService.js
```diff
@@ -60,11 +60,17 @@ class MapConfigService {
...
-    var proxyTiles = function(opt_proxy, config) {
-      return opt_proxy && (config.type === 'TileWMS' ||
-        (typeof config['use_proxy'] !== "undefined" && config['use_proxy'] === true));
-    };
-    if (proxyTiles(opt_proxy, config)) {
+    
+    // Always proxy if config's use_proxy = True
+    // otherwise use original conditional
+    if (opt_proxy && typeof config['use_proxy'] !== "undefined" && config['use_proxy'] === true) {
+      sourceObj.setTileLoadFunction((function() {
```
bartvde: instead of copying this function, can you put it into a shared function that can be used by both code blocks instead?

travislbrundage: Yeah, that makes sense. Will update.

PR #853, opened 2018-03-15
Fix for async source/update cycle.

theduckylittle:
When a source is created and needs an immediate update, an error
was being thrown as the source-creation is now done asynchronously.

Adding a promise to wrap the update of sources ensures the source
has been created before trying to update it.


src/components/map.js
```diff
@@ -659,8 +659,14 @@ export class Map extends React.Component {
     // check the sources diff
     const next_sources_version = getVersion(nextProps.map, SOURCE_VERSION_KEY);
     const next_layer_version = getVersion(nextProps.map, LAYER_VERSION_KEY);
+
+    // defeault to the source-configureation promise to being resolved.
```
bartvde: default

bartvde: configuration

PR #855, opened 2018-03-19
Set version in GetLegendGraphic request and use proxy

travislbrundage:
Proxy the legend graphic request if appropriate and add version to call

### DIFF
--- a/src/components/WMSLegend.js
+++ b/src/components/WMSLegend.js
@@ -129,8 +129,8 @@ class WMSLegend extends React.PureComponent {
     }
     url.set('query', query);
     // proxy the legend graphic
-    if (typeof layer.opt_proxy !== 'undefined' && typeof layer.use_proxy !== 'undefined' && layer.use_proxy === true) {
-      url = util.getProxiedUrl(url, layer.opt_proxy);
+    if (typeof source.opt_proxy !== 'undefined' && typeof source.use_proxy !== 'undefined' && source.use_proxy === true) {
+      url = util.getProxiedUrl(url, source.opt_proxy);
     }
     var legendUrl = (!this.state.dynamic && params.STYLES === undefined && params.SLD_BODY === undefined && layer.get('legendUrl')) ? layer.get('legendUrl') : url.toString();
     return (<img style={this.props.style} className={classNames('sdk-component wms-legend', this.props.className)} src={legendUrl} />);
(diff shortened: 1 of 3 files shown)


dakcarto:
Oops. I guess I shouldn't be reviewing changes on this repo.

src/components/WMSLegend.js
```diff
@@ -121,12 +121,17 @@ class WMSLegend extends React.PureComponent {
...
-      style: params.STYLES ? params.STYLES : ''
+      style: params.STYLES ? params.STYLES : '',
+      versoin: params.VERSION || '1.1.1'
     };
     if (params.SLD_BODY) {
       query.sld_body = params.SLD_BODY;
     }
     url.set('query', query);
+    // proxy the legend graphic
+    if (layer.opt_proxy !== 'undefined' && layer.use_proxy !== 'undefined' && layer.use_proxy === true) {
```
bartvde: this code does not make sense, you are probably missing typeof here?

I'm still doubting if tagging it onto the layer makes sense. The source makes more sense than maybe, but ideally that would not be needed. But I also can't think of a good way to prevent it right now.

src/services/MapConfigService.js
```diff
@@ -95,6 +95,10 @@ class MapConfigService {
     if (sourceConfig) {
       var source = this.generateSourceFromConfig(map, sourceConfig, opt_proxy, layerConfig.url, layerConfig.name);
       layer.setSource(source);
+      layer.use_proxy = sourceConfig.use_proxy;
+      if (opt_proxy !== 'undefined') {
```
bartvde: same here, typeof missing? Can we tag it to the source instead?

src/components/WMSLegend.js
```diff
@@ -121,12 +121,17 @@ class WMSLegend extends React.PureComponent {
       legend_options: options,
       layer: params.LAYERS,
       '_olSalt': this.state.salt,
-      style: params.STYLES ? params.STYLES : ''
+      style: params.STYLES ? params.STYLES : '',
+      versoin: params.VERSION || '1.1.1'
```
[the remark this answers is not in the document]
travislbrundage: Fixed the typo

src/components/WMSLegend.js
travislbrundage: Yeah I think I can move it to the source. I'm not sure what the best way is, but it needs to be somewhere for the legend graphic to recognize it, I just instinctively put it into the layer at first

bartvde: Okay let's just leave it on the layer. Can you verify that a save and reload of a map works fine?

src/services/MapConfigService.js
travislbrundage: Fixed

PR #857, opened 2018-03-20
Update the docs to say that no filter will remove everything

### LINKED ISSUE #840: MAP.REMOVE_FEATURES matches everything if filter is undefined
According to https://sdk.boundlessgeo.com/docs/module-actions_map.html `removeFeatures(sourceName, filter)` will remove everything that matches the provided filter. However, not prodiving a filter at all seems to result in a match function expression that kicks out everything: https://github.com/boundlessgeo/sdk/blob/master/src/reducers/map.js#L551

We're actually using `removeFeatures` currently because of that behavior, but the documentation does not suggest that it should work like that. I'd expect a missing filter to not match and delete anything.

bartvde:
closes #840

### DIFF
--- a/src/actions/map.js
+++ b/src/actions/map.js
@@ -247,7 +247,8 @@ export function addFeatures(sourceName, features, position = -1) {
   };
 }
 
-/** Remove features from a source on the map.
+/** Remove features from a source on the map. If no filter is provided, all features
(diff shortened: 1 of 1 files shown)


bartvde:
docs change only so merging

PR #859, opened 2018-03-20
Added error reporting for sources

theduckylittle:
Creates a new `sourceErrors` object on the
MapInfoReducer. `sourceErrors` are reset everytime
the map is moved.


__tests__/reducers/mapinfo.test.js
```diff
@@ -82,4 +83,39 @@ describe('mapinfo reducer', () => {
...
+    };
+
+    expect(reducer(state, action)).toEqual({
+      sourceErrors: {
+        test: true,
+      },
+    });
+  });
+
+  it('should errored sources', () => {
```
bartvde: clear missing here?

src/action-types.js
```diff
@@ -86,6 +86,7 @@ export const MAPINFO = {
   SET_RESOLUTION: 'MAPINFO.SET_RESOLUTION',
   SET_PROJECTION: 'MAPINFO.SET_PROJECTION',
   REQUEST_REDRAW: 'MAPINFO.REQUEST_REDRAW',
+  SET_SOURCE_ERROR: '',
```
bartvde: I guess this is not intentionally set as an empty string?

bartvde: also, is the clear source errors const missing?

src/actions/mapinfo.js
```diff
@@ -88,3 +88,26 @@ export function requestRedraw() {
     type: MAPINFO.REQUEST_REDRAW,
   };
 }
+
+/** Set an error for a source.
+ *
+ *  @param {String} srcName - The name of the source with the error
+ *
+ *  @returns {Object} Action object to request a map redraw.
```
bartvde: copy paste?

src/actions/mapinfo.js
```diff
@@ -88,3 +88,26 @@ export function requestRedraw() {
...
+export function setSourceError(srcName) {
+  return {
+    type: MAPINFO.SET_SOURCE_ERROR,
+    srcName,
+  };
+}
+
+/** Clear the source errors
+ *
+ *  @returns {Object} Action object to request a map redraw.
```
bartvde: same here, copy paste?

src/components/map.js
```diff
@@ -370,6 +370,8 @@ function getLoaderFunction(glSource, mapProjection, baseUrl) {
           this.addFeatures(GEOJSON_FORMAT.readFeatures(features, readFeatureOpt));
         }
       }).catch((error) => {
+        // use the event name tileloaderror for consistency.
+        this.dispatchEvent('tileloaderror');
```
bartvde: just checking, but this is the source scope here?

bartvde: I feel a little uncomfortable dispatching an event here on the openlayers object, but I also see in this function we don't have the map component scope to trigger the action directly

theduckylittle: This is necessary to "fake" the same behaviour that is available from the tile layers.  it also provides for easier testing as we have better test cases for the GeoJson type layers.

PR #863, opened 2018-03-22
Added error reporting to the layers list.

theduckylittle:
Groups and layers will now set a the CSS class `sdk-layer-error`
when there is an error with layer's source.


__tests__/components/layer-list.test.js
```diff
@@ -245,10 +250,24 @@ describe('test the LayerList component', () => {
...
     expect(wrapper.html()).toMatchSnapshot();
   });
 
+  it('should handle rendering an error on a group', () => {
+    setupGroups();
+
+    // add chaos to the wms layer.
+    store.dispatch(setSourceError('wms'));
+    const wrapper = mount(<Provider store={store}><SdkLayerList enableDD={false} /></Provider>);
+    expect(wrapper.html().indexOf('sdk-layer-error') > 0).toBe(true);
```
bartvde: minor but I would personally test this with: ```wrapper.find('.sdk-layer-error').to.have.length(1);```

PR #865, opened 2018-03-23
Fixes for async sources and refresh single source

bartvde:
SDK-939
SDK-922

The issues got introduced in #853 
It also adds the ability of force refreshing a single (MVT) source


bartvde:
Unit test added, that does not work without the fix, so ready for review now

bartvde:
Even with the change in the mapinfo props, it is still calling componentWillReceiveProps on every mousemove

bartvde:
Okay that seems to be caused by the castling MapScreen component that would pass on all state

bartvde:
@theduckylittle I added yet another commit to this branch, which allows to force refresh a single source. I also found out the new cache invalidation stuff you added was not implemented for MVT, so I added it

bartvde:
Thanks for the review on the plane @theduckylittle

PR #867, opened 2018-03-26
Add pipeline build for examples deployer

nstires-boundless:
Update header, address false fail on info


bartvde:
@nstires-boundless I am going to see if I can integrate this into our existing Jenkins pipeline instead, so we don't need 2 separate ones. Will keep you posted.

PR #870, opened 2018-03-30
Added Additional Documentation Checks.

theduckylittle:
1. Fixed issue with prototype-plugin which was throwing
   an exception on checking for "@ignore".
2. Added new script which will look for exports with missing
   documentation.
3. Fixed up missing docs found by #2.
4. Found a broken snap-shot test with the zoom slider where
   the attributes of the input had shuffled themselves around.
5. Fixes a small bug with clearSourceErrors: specifically when the map is "unconnected" clearSourceErrors had no default behaviour.


docs/doc_checker.js
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env node
+
+/** Small script to check to see if the exported functions of
+ *  a file has docs.
```
bartvde: has -> have

src/components/mapboxgl.js
```diff
@@ -65,6 +65,8 @@ function getVersion(obj, key) {
   return obj.metadata[key];
 }
 
+/** MapBoxGL based renderer class.
```
bartvde: Mapbox instead of MapBox (they seem sensitive about their spelling ;-))

src/components/mapboxgl.js
```diff
@@ -488,6 +490,12 @@ function mapStateToProps(state) {
   };
 }
 
+/** Get the extent for the map.
+ *
+ *  @param {MapBoxGLMap} map - The MapBoxGL Map instance.
```
bartvde: Mapbox

src/constants.js
```diff
@@ -68,6 +68,8 @@ INTERACTIONS.measuring = [
   INTERACTIONS.measure_polygon,
 ];
 
+/** Export all the const's in a conventient Object.
```
bartvde: convenient

bartvde:
@theduckylittle should we have a script in our package.json to run this? Should it run as pretest or something?

theduckylittle:
Cleaned up the comments and added the checker to pretest!

PR #880, opened 2018-04-17
Fix scope issue in tileloaderror listener

bartvde:
selection_570


bartvde:
I decided not to use the optional this argument, since this is gone in ol  5

PR #881, opened 2018-04-19
Make stopEvent configurable

bartvde:
So that people can have full control over this setting.


bartvde:
small config change, so merging

PR #883, opened 2018-04-20
Add load events to Map component

bartvde:
SDK-964


bartvde:
it seems the tileloaderror path is not getting triggered, seems to be conflicting with the other tileloaderror listener somehow, need to investigate a bit more

bartvde:
false alarm, it was just the testcase prop not having setSourceError, updated the test case

theduckylittle:
I had to sleep on this one a little bit.  The combination of call-backs and ref-counters took me a second to resolve how this best plays into a redux-based environment.  I believe the callback is fine but we might want to default the prop to dispatch a set of actions to the MapInfo reducer in a follow-up PR.

PR #884, opened 2018-04-20
Add a prop to make a map not interactive

bartvde:
If set to false, the map will not have any interactions in it, and hence will not response to any mouse or keyboard events


bartvde:
Minor change and tested by @CarlyLanglois so merging this one

PR #887, opened 2018-04-24
Use addOsmSource

bartvde:
see discussion in #886


CarlyLanglois:
@bartvde fwiw, i considered this too but i liked that the more generic addSource gives the user more of a demonstration of what adding a source looks like

bartvde:
agreed, it might be appropriate for a next-level tutorial (I think @theduckylittle is working on that), IMHO getting started should be as simple as possible.

PR #889, opened 2018-04-24
Refactor our examples

### LINKED ISSUE #888: Basic example is getting too complicated
Need to split out some functionality into separate examples

bartvde:
fixes #888


bartvde:
example changes only so merging without review

PR #893, opened 2018-05-17
Use zIndex on all layer types

bartvde:
SDK-984


bartvde:
Minor change only so merging without formal review

PR #897, opened 2018-05-22
Handle promise rejection from ol-mapbox-style applyStyle

bartvde:
ol-mapbox-style applyStyle returns a promise that can get rejected if the sprite cannot be loaded, and we were not handling this


bartvde:
Minor change, merging

PR #899, opened 2018-05-28
Fix async render issues

bartvde:
SDK-1018

Trying to tackle the async render related issues in SDK as discussed here: https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

- [x] History
- [x] Legend
- [x] Map
- [x] MapboxGL

### DIFF
--- a/__tests__/components/legend.test.js
+++ b/__tests__/components/legend.test.js
@@ -280,25 +280,6 @@ describe('test the Legend component', () => {
     expect(legend).toEqual(null);
   });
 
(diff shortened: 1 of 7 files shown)


bartvde:
shouldComponentReceiveProps never actually seemed to do anything in a real life example, I tested in the paint-change example and when selecting the same color, it would still return true and shouldComponentUpdate was the one preventing the actual update.

bartvde:

PR #861, opened 2018-03-21
Upgrade to ol5

bartvde:
Now ready for review

PR #900, opened 2018-05-28
Fix undefined refs which kills Mapbox libs, have default zoom and center

bartvde:
This is taken from #861 since it might take a while before we upgrade, and it fixes actual issues in the current code base

### DIFF
--- a/__tests__/components/map.test.js
+++ b/__tests__/components/map.test.js
@@ -536,8 +536,8 @@ describe('Map component', () => {
     const view = map.getView();
     // default values should get set
     expect(view.getRotation()).toBe(0);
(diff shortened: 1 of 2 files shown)


bartvde:
@theduckylittle any chance you can check performance on your end (no rush)? For me it is really slow in the openlayers renderer. It was broken before this change.

bartvde:
@ahocevar it looks like we have a performance issue again mapbox styles in the OL renderer in SDK, any pointers as to what might have broken?

bartvde:
ah false alarm guys, it seems my Chrome instance was borked, it works fine in a fresh instance

PR #902, opened 2018-06-01
Added Support for Headers with Tile Fetches

theduckylittle:
1. Checks for an Authorization header in fetchOptions.
2. When present, uses an async loader to get the data
   using fetch, convert it to a BLOB and then set the Tile's
   image to that URL.

### DIFF
--- a/src/components/map.js
+++ b/src/components/map.js
@@ -125,15 +125,39 @@ function getVersion(obj, key) {
   return obj.metadata[key];
 }
 
+/**
+ * Load a tile with full fetchOptions.
+ *
+ * This is only used when the Authorization header has been set
+ * in the fetchOptions.
+ *
+ * @param {Object} fetchOptions Options to use for fetch calls.
+ *
+ * @returns {Function} Tile loading function which uses fetch.
+ */
+function authTileLoader(fetchOptions) {
+  return function(tile, src) {
+    fetch(src, fetchOptions)
+      .then(r => r.blob())
+      .then((imgData) => {
+        tile.getImage().src = URL.createObjectURL(imgData);
+      })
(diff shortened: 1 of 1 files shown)


src/components/map.js
```diff
@@ -125,15 +125,39 @@ function getVersion(obj, key) {
   return obj.metadata[key];
 }
 
+/**
+ * Load a tile with full fetchptions.
```
bartvde: fetchOptions

src/components/map.js
```diff
@@ -125,15 +125,39 @@ function getVersion(obj, key) {
...
+ * @param {Object} fetchOptions Options to use for fetch calls.
+ *
+ * @returns {Function} Tile loading function which uses fetch.
+ */
+function authTileLoader(fetchOptions) {
+  return function(tile, src) {
+    fetch(src, fetchOptions)
+      .then(r => r.blob())
+      .then((imgData) => {
+        tile.getImage().src = URL.createObjectURL(imgData);
```
bartvde: we probably need a polyfill for URL?

src/components/map.js
```diff
@@ -125,15 +125,39 @@ function getVersion(obj, key) {
...
+      })
+      .catch(() => {
+        console.error('Error fetchimg image at:', this.props.src);
```
bartvde: fetching

bartvde: this.props.src or simply src?

src/components/map.js
theduckylittle: Do we? https://caniuse.com/#feat=bloburls

bartvde: ah right, seems we don't need it

PR #905, opened 2018-06-04
Add new example with more than 1 map

### LINKED ISSUE #904: multiple maps?
Hi! Thanks for your work!

This is more a question than an issue. 
I am trying to make two maps on one page, but can't think of a way to achieve this with boundless sdk. 
 Can I get direction for this?

bartvde:
closes #904


bartvde:
example only so merging

PR #906, opened 2018-06-05
Add support for fetch options to VectorTiles

theduckylittle:
The previous Auth0 support PR only worked with Images
this version adds support to vector tiles.

### DIFF
--- a/src/components/map.js
+++ b/src/components/map.js
@@ -148,6 +148,35 @@ function authTileLoader(fetchOptions) {
   };
 }
 
+/**
+ * Load a vector tile with full fetchOptions
+ * This is only used when the Authorization header has been set
+ * in the fetchOptions.
+ *
+ * @param {Object} fetchOptions Options to use for fetch calls.
+ *
+ * @returns {Function} Tile loading function which uses fetch.
+ */
+function authVectorTileLoader(fetchOptions) {
+  return function(tile, url) {
+    const loader = () => {
+      fetch(url, fetchOptions)
+        .then(r => r.arrayBuffer())
+        .then((source) => {
+          const format = tile.getFormat();
+          tile.setProjection(format.readProjection(source));
+          tile.setFeatures(format.readFeatures(source));
+          tile.setExtent(format.getLastExtent());
+        })
+        .catch((err) => {
+          tile.onError();
+        });
+    };
+
+    tile.setLoader(loader);
+  };
+}
+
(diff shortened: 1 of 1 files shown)


src/components/map.js
```diff
@@ -148,6 +148,33 @@ function authTileLoader(fetchOptions) {
...
+ * @returns {Function} Tile loading function which uses fetch.
+ */
+function authVectorTileLoader(fetchOptions) {
+  return function(tile, url) {
+    const loader = () => {
+      fetch(url, fetchOptions)
+        .then(r => r.arrayBuffer())
+        .then((source) => {
+          const format = tile.getFormat();
+          tile.onLoad(format.readFeatures(source), format.readProjection(source), format.getLastExtent());
```
bartvde: this (```onLoad```) looks to be outside of the API? Probably cleaner to use setFeatures instead like here: https://openlayers.org/en/latest/apidoc/ol.source.VectorTile.html ?

bartvde: instead of the internal ``onError``` we should likely trigger an event instead?

bartvde: I know this is less important with the ```ol``` package which exports everything, but if there is something in the API to do the same I would prefer to get that route, to avoid breakages in the future.

theduckylittle: I would agree but the VectorTile source actually uses the VectorImageTile class and the "stock" loader from OL looks like this:
```
_ol_VectorImageTile_.defaultLoadFunction = function(tile, url) {
  var loader = _ol_featureloader_.loadFeaturesXhr(
      url, tile.getFormat(), tile.onLoad.bind(tile), tile.onError.bind(tile));

  tile.setLoader(loader);
};
```

So I was trying to mimic what was in the stock code because as far as I can see there _is not_ a setFeatures function as far as I can see.

bartvde: ping @ahocevar any thoughts on this?

ahocevar: An example tile load function is listed in the [API docs](https://openlayers.org/en/latest/apidoc/ol.source.VectorTile.html), but it is incorrect. I'm going to update that. A correct minimal `tileLoadFunction` would look like this:
```js
function(tile, url) {
  tile.setLoader(function() {
    const format = tile.getFormat();
    fetch(url).then(function(response) {
      return response.arrayBuffer();
    }).then(function(data) {
      tile.setProjection(format.readProjection(data)),
      tile.setFeatures(format.readFeatures(data, {
        extent: format.getLastExtent()
      }));
    });
  });
}
```

PR #916, opened 2018-07-02
Allow specification of the vendor-specific param TILED

bartvde:
MBSE-105


bartvde:
@theduckylittle PR updated per your great suggestion

PR #921, opened 2018-07-13
fixing typo in filter example

mrcnc:


wnordmann:
I know I made this example, so I likely made the typo too.  It looks like something I would do

PR #925, opened 2018-08-03
Use redux-saga instead of redux-thunk

bartvde:
SDK-1038


bartvde:

bartvde:
@theduckylittle just realized I forgot to remove the thunk action creator, new PR upcoming

PR #928, opened 2018-08-10
Fix examples in IE11

bartvde:
fixes SDK-1043


theduckylittle:
Would a simple change be to tweak the config to just include the polyfill as an entry point like so:

```
    entry: [
        'babel-polyfill',
        '...
    ],
```

bartvde:
I will give that a try @theduckylittle

bartvde:
PR updated @theduckylittle

PR #935, opened 2018-09-12
Fix typo in sourceName for selection events

bartvde:
The drawing prop is called ```sourceName``` not ```sourcename```


bartvde:
non functional change so merging

PR #948, opened 2018-10-17
Bug: Do not set HTML context on the component.

theduckylittle:
According to the react-dnd docs the context should
be set for the entire application.  While developing an app
I found that I could not add other DND targets or sources because the
context was limited to only the layers list.


theduckylittle:
Examples have been updated!

PR #954, opened 2018-12-20
Fix the getting started guide

bartvde:
With react-redux version 6, our getting started guide broke

Also ejecting is not needed anymore with the latest create-react-app, so that section was removed.

I also moved advanced stuff to the end of the tutorial to not complicate things for the novice, it's already complicated as it stands


bartvde:
doc change only so merging without review

PR #955, opened 2018-12-24
[WIP] Upgrade webpack

bartvde:
refs SDK-1079

Since webpack-extract-text-plugin is not compatible, we needed to replace it with mini-css-extract-plugin

https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/701

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,6 @@ After this, run the following commands:
 
 `npm install` - install libraries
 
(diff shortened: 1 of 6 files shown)


bartvde:
It should be possible to have a single file for the css as well, see: https://webpack.js.org/plugins/mini-css-extract-plugin/#extracting-all-css-in-a-single-file

Issue #791, opened 2018-01-10
Add Support for `MultiPolygon`


milafrerichs:
Can we add support for drawing `MultiPolygons` ?

That would be awesome.
Thanks.
I can do a PR if you want.

bartvde:
Sure, PR welcome ofcourse

PR #961, opened 2019-01-11
Adding in a new example

wnordmann:
Started to check on #791 turns out support exists so I wanted to make an example to prove it.  Example is base on Openlayers example https://openlayers.org/en/latest/examples/geojson.html .  Found that Openlayers allows unclosed polygons while SDK does not.  Also noticed SDK doesn't handle Points inside of GeometryCollections.

### DIFF
--- a/examples/multigeometry/app.css
+++ b/examples/multigeometry/app.css
@@ -1,0 +1,80 @@
+/* Demo how to over ride the default color of the buttons */
+
+button.sdk-btn.blue {
+    background-color: #29728d;
+}
+
+#controls form {
+  font-size: .85rem;
+}
+
+.mapName {
+  font-weight: 400;
+  line-height: 1.25;
+  text-align: left;
+  white-space: nowrap;
+  vertical-align: middle;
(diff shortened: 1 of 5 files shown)


bartvde:
the ticket was about the draw tools, so the user needs to be able to draw a multipolygon using the draw interaction

wnordmann:
Drawing a multipolygon, that makes a lot more sense.  I'll make a new Jira ticket.

theduckylittle:
Chatted with @wnordmann , I'm going to approve this and he will have a follow up PR with the drawing tools as an example.

bartvde:
Please note it doesn't work well in the Mapbox GL renderer

screenshot from 2019-01-16 08-37-39

wnordmann:
Mapbox GL render only likes CRS84 / EPSG:4326 and doesn't like EPSG:3857.  Easy enough to change the sample geojson.

wnordmann:
Mapbox render is addressed in https://github.com/boundlessgeo/sdk/pull/962

PR #962, opened 2019-01-16
Changing the EPSG code on the example to work with mapbox

wnordmann:
Mapbox GL render doesn't like EPSG 3857, changed example json to use CRS:84 / EPSG 4326


bartvde:
looks good thanks for addressing

PR #963, opened 2019-02-01
Update name of app

travislbrundage:
Updating the name of the app on the `1.x` branch to be consistent with `master` and how it is referenced in [`geonode-client`](https://github.com/GeoNode/geonode-client/blob/master/package.json#L39)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ Playback.css
 sdk.css
 Geocoding.css
 FeatureTable.css
+.idea/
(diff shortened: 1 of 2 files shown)


bartvde:
@travislbrundage this shouldn't be needed since the script takes care of it: https://github.com/boundlessgeo/sdk/blob/1.x/scripts/copy-files.js#L60

travislbrundage:
Good to know! Thanks!

